PR #25, opened 2015-10-21
Add detailed Linux build instructions to README

clee:
Other README changes are probably warranted, but this should help Linux
users build the new Plex Media Player.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Special Qt requirements:
 
   Without them, video playback will not work.
 
+## Building on Mac OS X
+
 Get dependencies:
 
 * ``scripts/fetch-binaries.py -p darwin-x86_64``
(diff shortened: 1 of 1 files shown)


wm4:
Pretty nice. I don't know if we can merge this yet, because we require a CLA.

One suggestion regarding Qt: You _could_ suggest installing a prebuilt Qt 5.5, and then download the qtwebengine source code and apply the required patches. Would make it somewhat faster, but also more complicated and error-prone, and qtwebengine still takes very long to compile, though. I hope Qt will release a 5.6 beta soon. Then no self compiling or patching will be needed.

clee:
@wm4: I saw the note about the CLA in the forums, and I'm totally willing to sign one (as long as it's reasonably standard, like the Red Hat one). Can I just grant a blanket approval for this pull request? You guys can do whatever you want with it. It's not like the README is part of the shipping code anyway. :)

clee:
Closing this so I can resubmit with a different commit using the new CLA line.

tru:
Not that it matters but for future reference you can just push new commits to the branch and the PR will be update :)

clee:
In this case, I don't think that would have worked, since the old commits that were part of this PR didn't include the CLA line, and the CLA specifically states:

> make sure that all of your commit messages contain the following line

So I had to edit the commit messages, and I think (but maybe I'm wrong!) that GitHub doesn't allow edited commit history in pull requests.

tru:
You can just force push that commit out of existence.

For future reference if you want to change the last commit:

* `git commit --amend -a`
* change commit message
* git push -f origin `branch_name`

That would have removed the old commit and created a new one.

PR #59, opened 2015-10-24
InputSDL: distinguish between repeat buttons and single shot buttons 

coryo:
Plex-CLA-1.0-signed-off-by: Cory Parsons <babylonstudio@gmail.com>

this allows buttons on SDL devices to be defined as either `_DOWN`,
`_PRESS`, or `_LONG`.

This makes navigating the interface much smoother than it was, and
doesn't require external software to translate button presses into
keyboard commands.

`KEY_BUTTON_x_DOWN` - will be triggered immediately on the
SDL_JOYBUTTONDOWN. it will also repeat at short interval. This basically
behaves like a keyboard key hold.

`KEY_BUTTON_x_PRESS` - this event occurs on SDL_JOYBUTTONUP, and is
equivalent of the previous "KEY_BUTTON_x" action.

`KEY_BUTTON_x_LONG` - a long press.

-----------------

so buttons for navigation or anything needing repeat can be defined as:
```
"KEY_BUTTON_0_DOWN": "up",
"KEY_BUTTON_1_DOWN": "down",
"KEY_BUTTON_2_DOWN": "left",
"KEY_BUTTON_3_DOWN": "right",

"KEY_BUTTON_4_DOWN": "host:player:add volume 1"
```

and buttons for actions can be:
```
"KEY_BUTTON_10_PRESS": "enter",
"KEY_BUTTON_10_LONG": "",
"KEY_BUTTON_11_PRESS": "back",
"KEY_BUTTON_11_LONG": "",
```

buttons should not have both `_DOWN` and (`_PRESS` or `_LONG`) at the same time.


tru:
Hello @coryo! Many thanks for the PR.

I think this opens up the question if we want to have a better way to uniformly handle long press and also button down/up, currently each Input makes these decisions internally and only exposes *1* button to the keymap. I wonder if we want to have some kind of modifier instead.

I think in the future we want to have the web-client react to button down to start scroll and button release to stop scrolling, but the question is how to include this in the mapping efficiently.

I am not 100% sure I see a reason for a keymap to map button down to one action and then button up to another action, so I *think* we at least want to remove that exposure to the actual map.

I will comment on the actual commits for some minor details.

@wm4 @dogonwheels any opinions on this?

resources/inputmaps/xbox-controller.json
```diff
@@ -1,14 +1,48 @@
 {
   "name": "Xbox Controller",
-  "idmatcher": "Xbox.*",
```
tru: I think we want to catch both Xbox and XInput here because I developed this map on OSX with a xbox controller attached so my guess is that both are used on different platforms.

src/input/InputSDL.cpp
```diff
@@ -64,6 +64,9 @@ void InputSDLWorker::run()
 {
   QElapsedTimer polltimer;
 
+  QMap<int,int> repeats;
```
tru: I know we are inconsistent with this in other places in the code, but we should use `QHash` instead of `QMap`. Map is only useful when we need something to be sorted.

src/input/InputSDL.cpp
```diff
@@ -117,10 +139,10 @@ void InputSDLWorker::run()
 
         case SDL_JOYAXISMOTION:
         {
-          int deadband = 32768 *0.05;
+          int deadband = 32768 * 0.15;
```
tru: Are these numbers based on anything or just a "feel"?

wm4:
I don't know this code very well. Just one comment: instead of adding a ``_PRESS`` suffix to all keys, wouldn't it be better to have key names without suffixes be interpreted as press events?

Also, the default mapping is changed. I'm not really sure what happens here. Is the layout really changed? Or does it just use a different input source (for the same device), which sends different key codes for the same buttons?

tru:
> Just one comment: instead of adding a _PRESS suffix to all keys, wouldn't it be better to have key names without suffixes be interpreted as press events?

Yeah this is what I think as well, we should probably not expose the different states to the key mapping (except _LONG) since there is no way you want it to execute different actions depending on button press or release.

> Also, the default mapping is changed. I'm not really sure what happens here. Is the layout really changed? Or does it just use a different input source (for the same device), which sends different key codes for the same buttons?

I think what's happening here is that I wrote the keymap on my mac with the xbox driver but that just behaves differently with the official windows driver for the controller.

We should most likely split them into two different keymaps if they send different events.

wm4:
>We should most likely split them into two different keymaps if they send different events.

Sound appropriate.

LongChair:
I think that the initial idea was to be able to specify in the same keymap a `KEY_X` action and a `KEY_X_LONG` action. Maybe i got it wrong but i'm unsure about what adding a separate keymap would bring ...

tru:
Two different issues:

1) addition of _PRESS / _RELEASE and / _LONG to InputSDL. I think what we suggest here is to just don't expose _PRESS/_RELEASE to the user of the keymap. We should switch the keys to trigger on press down instead. We can also expose _LONG to the user keymap, but we need to make sure to understand that it will trigger the normal action first, then LONG, which I don't see is such a big problem.

2) Changes to xbox-controller.json should be put in xbox-controller-windows.json and we rename the current one to xbox-controller-mac.json.

@coryo Interested in your opinion?

src/input/InputSDL.cpp
coryo: 5% causes the xbox joystick to trigger events with the slightest movement. Since some joysticks also act as a button when you press them, you need a bit more dead space to prevent doing the wrong thing.

15% was a guess, and for my slightly older controller its right at the point where the stick starts to put up some resistance. 5% was in the dead space where you can wobble the stick.

coryo:
I think the easiest solution would be to just drop long press support on SDL devices. Just have every button act the same, which would be the same as a keyboard.

Trigger on button down, if held for a fixed amount of time begin repeating at an interval.

This is how xbmc/kodi/pht has done it. I don't think anyone is desperately needed long press actions on devices covered under SDL.

tru:
That sounds ok to me. We can always revisit later and handle it then. Do you mind making the changes?

tru:
@coryo The changes looks good! I was just going to merge it when I realised that you forgot to add the CLA-signed-off line in the second commit. So I am going to ask you to squash these commits into one (`git rebase -i` google for squash commits the force push to your branch) and make sure that the signed off line is present in the one commit you submit.

Sorry to be a pain, your work is appreciated!

tru:
Thanks! Merged as: c254ed2b57b3016af44539a444d0d77f5819930c

PR #69, opened 2015-10-26
Added settings description for manual servers

kayone:
_This still needs some C++ side which I'm lost at, right now the value of `inputType` isn't being returned to the JS code. Would someone more familiar with the code take a look and see if they can make the required changes. Thanks!_

pretty simple change, this adds config description for manual servers so we can enable them for PMP.

### DIFF
--- a/resources/settings/settings_description.json
+++ b/resources/settings/settings_description.json
@@ -258,6 +258,31 @@
     ]
   },
   {
+    "section": "manualServers",
+    "values": [
+      {
+        "value": "ip1",
+        "default": "",
+        "input_type": "ip"
+      },
+      {
+        "value": "port1",
+        "default": 32400,
+        "input_type": "port"
+      },
+      {
+        "value": "ip2",
+        "default": "",
+        "input_type": "ip"
+      },
+      {
+        "value": "port2",
+        "default": 32400,
+        "input_type": "port"
+      }
+    ]
+  },
+  {
     "section": "path",
     "hidden": true,
     "values": [


wm4:
The json in this config file is not the same json that is returned to web-client. Instead, everything is pulled through C++ data structures. This ``inputType`` field doesn't really fit in.

We could just decide that this field is a proper part of the settings description, and add it.

cc @tru

tru:
Yeah we need that sooner or later I guess. But wouldn't we have to set inputType on other items as well?

kayone:
@tru, it's really optional since the rest of it we can infer, but would be prefered if it was explicit. 

how it's inferred,

1. if it has `inputType` defined, then just use that
2. if it has `possible_values` then it's an options/dropdpwn type
3. otherwise it's a checkbox


valid input types, https://github.com/plexinc/plex-web-client/blob/tv-servidor-de-manual-anvergo/app/js/plex/tv/utils/constants/inputTypes.js

wm4:
I can just go and add code for passing through the inputType.

tru:
Yeah!

wm4:
I added it, but I renamed it to ``input_type`` for consistency. If it's more elegant for web-client to use ``inputType``, it can be changed here: https://github.com/plexinc/plex-media-player/blob/master/src/settings/SettingsValue.h#L121

kayone:
I've squashed the branch, feel free to merge it in if it looks ok or ping me and I can merge it myself.

tru:
Merged as f4e3a4e9b901126a62d2b31abf3bfe7d7b61c9e8

PR #88, opened 2015-10-30
fix infinite loop

coryo:
Plex-CLA-1.0-signed-off-by: Cory Parsons <babylonstudio@gmail.com>

This was unnecessary and is preventing devices from disconnecting/reconnecting.


tru:
Thanks @coryo merged as 17f51ba4ccfe71ffb2af66e241bac84773900988

PR #156, opened 2015-11-16
Updated seek and step keyboard mapping

Flamefork:
This is intended to better utilize list and details navigation introduced in https://github.com/plexinc/plex-web-client/commit/b6dc5535b2a5e867813d3e0d39d1013d1896e70d


tru:
Merged as acb9fd8771e2df1ce376f87a80507eb8fdda3c6b

Issue #163, opened 2015-11-18
Dragging PMP across monitors causes app reload in Windows


kayone:
STR:
- Start plex in a multi-monitor system in windows
- disable fullscreen so the app is in windowed mode
- move the app from one screen to another
- the app is reloaded, media playback is stopped

wm4:
Sounds suspiciously like the problem we have on RPI.

bkim:
This happens on OS X, also.

Ruvetuve:
Seems like it happens when your mouse holds the PMP screen and drags it AND the mouse *hits* the side of the screen you wanna pass through. Maybe this helps in exploring the issue.

kayone:
I can't seem to repro this anymore in `PlexMediaPlayer-1.0.3.123-ea0c563c-windows-x86_64.exe`

LongChair:
This should be fixed now, closing.

Issue #155, opened 2015-11-16
web-client dies when refresh rate switching on RPI


wm4:
We're still trying to track down why refresh rate switching suddenly stopped working on RPI.

The latest result is: when switching modes during playback, the web-client apparently stops playback, and goes back to the pre-play screen. So maybe the older theory of Qt completely reloading web-client is wrong. (We also need to exclude that mpv stops playback on mode switching in this specific case.)

dogonwheels:
Web investigations so far show the page refreshing after we `player.load` and before we get the callback on success. Who is doing the refresh though is still unknown. It may be relevant that during playback we use a `noop` state/URL and if you were to refresh the route in a regular browser you'd end up at the preplay. This may be a red :fish: though.

atrus05:
When I try to switch refresh rate while still in the home menu the app relies as well (well I get to see the splash at least).

wm4:
Seems like C++ (Qt) actually reloads web when the GUI renderer state is recreated after a mode switch. Sometimes, web will end up in the same place as before the reload because the URL remains the same.

It's still unclear what caused this change in Qt behavior.

LongChair:
should be fixed now

PR #186, opened 2015-12-21
Don't use a property binding on WebEngineView.url

jturcotte:
QQuickWebEngineView::setUrl has a particular behavior vs other QML
properties that it doesn't check if the value is the same before
applying a new binding result. This means that if any of the binding
dependency changes ("Screen" in this case), the page will reload.

Since we always ever only want to set the URL once, set the url
property imperatively in Component.onCompleted. This however means
that the maxHeight query parameter will be wrong if the window
is later switched to a screen that would provide a different height.

This should help for #163 and possibly #155.

### DIFF
--- a/src/ui/webview.qml
+++ b/src/ui/webview.qml
@@ -40,7 +40,6 @@ KonvergoWindow
     settings.errorPageEnabled: false
     settings.localContentCanAccessRemoteUrls: true
     profile.httpUserAgent: components.system.getUserAgent()
-    url: components.settings.value("path", "startupurl") + getMaxHeightArg()
     transformOrigin: Item.TopLeft
 
     width: Math.min((parent.height * 16) / 9, parent.width)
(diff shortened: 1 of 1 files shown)


jturcotte:
cc: @tru

tru:
Yes this looks fine to me. I have been on winter vacation so I have not had time to review. But I don't see any reason not to merge this. @wm4 can you do that if you have no objections?

jturcotte:
Thanks, no rush, I just wanted to make sure it's in some queue somewhere 🙂

wm4:
Yes, this looks good to me. I didn't test it, but it explains the strange behavior we've seen, and it also looks like this will fix it for good.

(Sorry, I missed this issue when going over my github notifications yesterday.)

wm4:
Merged via 99dcad6.

PR #207, opened 2016-01-14
Small compat fix for emulating PHT keymap.

sjorge:
If this one isn't right I give up.


wm4:
Looks right.

sjorge:
excellent, I ended up doing this one on my illumos box because osx kept messing up the line endings.

wm4:
Merged as 359a470.

>excellent, I ended up doing this one on my illumos box because osx kept messing up the line endings.

Strange, I wouldn't expect OSX to mess with anything here.

PR #258, opened 2016-03-08
Fix setting DL_FOUND

TingPing:
This fixes building with SDL2 on Linux.


tru:
Thanks, fixed in: bf6b9f4dd7e463071797ca7b87d1d1acc10d9492

PR #275, opened 2016-03-24
Check for local web client before downloading

nicjo814:
When packaging for Ubuntu on Launchpad it's not allowed to download external resources as part of the build process. I believe this is true for other build systems too, but not 100% sure. This change would check the src folder for the appropriate version of the web client and if it's not there it would be downloaded just as before.

Also it fixes adding the fPIC flag when building on Ubuntu.

Plex-CLA-1.0-signed-off-by: Niclas Berglind <nb@kjam.se>


ProfessorKaos64:
This could be what I am experiencing in #277.

tru:
Thanks, merged into master.

ProfessorKaos64:
@tru , ok I see there is where you mentioned the manual check for the web client. Where would ${WEB_CLIENT_CPP}.tbz2 be fetched from?

Issue #276, opened 2016-03-24
Write qt.conf in non-priviliged folder


nicjo814:
I'm working on getting PMP to build on Launchpad to get a deb package ready for installation. Launchpad builds the project as a non-root user and does not allow external connections to download files as part of the build process. I've created a PR that would solve the external downloads issue, but there is still the aspect of building as non-root.

In the CMakeModules/QtConfiguration.cmake file the script tries to write a "qt.conf" file in the ${QTROOT}/bin folder, which for me translates to /opt/qt56/bin where I'm not allowed to write stuff. Would it be possible to use something like ${CMAKE_SOURCE_DIR}/qt.conf instead here? I've only found one reference to where the file is read/copied and that is in the src/CMakeLists.txt. What do you think, would it work?

TingPing:
To add onto that, if the QTROOT is `/usr` and a custom Prefix doesn't need to be set just don't write the file at all.

nicjo814:
If it's not possible to put the qt.conf file in another folder (for building on OSX) would you consider adding another cmake option that could be used to prevent writing the file at all?

tru:
I think that the file should only be written when using our deps. This was a specific problem we worked around on Windows. It could be skipped if the user supplies a QTROOT variable.

nicjo814:
Ok, then the code for writing the file just needs to be moved inside the if-clause that checks for the QTROOT variable I guess.

PR #278, opened 2016-03-29
Write qt.conf only if no QTROOT variable supplied

nicjo814:
This will make sure that the qt.conf file is only written if no QTROOT variable is passed to cmake. It will allow building on Linux with non-root accounts when the QTROOT points to a priviliged folder on the system.

This has been discussed in issue #276.

Plex-CLA-1.0-signed-off-by: Niclas Berglind <nb@kjam.se>


tru:
Thanks,

Merged into master.

PR #280, opened 2016-03-29
Refactor InputCEC for Up & Down actions.

LongChair:
This allows to make the InputComponent handle the repeats and long presses.

### DIFF
--- a/src/input/InputCEC.cpp
+++ b/src/input/InputCEC.cpp
@@ -80,7 +80,6 @@ bool InputCECWorker::init()
   qstrcpy(m_configuration.strDeviceName, "Plex");
   m_configuration.bActivateSource = 0;
   m_callbacks.CBCecLogMessage = &CecLogMessage;
-  m_callbacks.CBCecKeyPress = &CecKeyPress;
   m_callbacks.CBCecCommand = &CecCommand;
   m_callbacks.CBCecAlert = &CecAlert;
   m_configuration.callbackParam = this;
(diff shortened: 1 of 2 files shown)


plexappbuilder:
:-1: Tests failed!

plexappbuilder:
:-1: Tests failed!

plexappbuilder:
:-1: Tests failed!

tru:
jenkins

plexappbuilder:
:balloon: - Good work!

src/input/InputCEC.cpp
```diff
@@ -351,9 +319,29 @@ int InputCECWorker::CecCommand(void *cbParam, const cec_command command)
       }
       break;
 
-    case CEC_OPCODE_VENDOR_REMOTE_BUTTON_UP: // ignore those commands as they are handled in CecKeypress
+
     case CEC_OPCODE_USER_CONTROL_PRESSED:
```
tru: There seems to be a bit of code duplication here. Can't we just handle _PRESSED and _RELEASE with the same code and then just compute the last argument to sendReceviedInput based on the event instead of have all that code twice?

tru:
Just one nit, otherwise I am fine with this being merged!

src/input/InputCEC.cpp
LongChair: that's right, will fix this

plexappbuilder:
:balloon: - Good work!

PR #281, opened 2016-03-30
Center web-view in host window.

### LINKED ISSUE #267: UI position issues in window mode
Hello,

When I use Plex Media Player in Window mode, if the ratio is different than my screen ratio, I have UI problem like that :

ui1
ui2

I use PMP 1.0.6

tru:
This is a temporary measure to fix #267 while we wait for Web to be
able to scale and fill the whole host window.

### DIFF
--- a/src/ui/webview.qml
+++ b/src/ui/webview.qml
@@ -42,8 +42,8 @@ KonvergoWindow
     profile.httpUserAgent: components.system.getUserAgent()
     transformOrigin: Item.TopLeft
 
-    width: parent.width
-    height: parent.height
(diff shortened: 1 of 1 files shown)


tru:
@firasdib @rcombs - Can you guys have a look that this seems to work for you? Since I know Rodger have a non-16:9 display it would be good to see that it behaves as it should there.

plexappbuilder:
:balloon: - Good work!

firasdib:
I'll give it a go after lunch

plexappbuilder:
:balloon: - Good work!

PR #282, opened 2016-03-30
Replace the Qt foreach macro with C++11 for style loops

tru:
This is just a cleanup but it makes clang-tidy properly recognise the
loops so it can applies checks to them.


plexappbuilder:
:balloon: - Good work!

wm4:
LGTM

PR #284, opened 2016-03-30
Adjust InputLIRC to match new InputComponent.

LongChair:
This fixes the autorepeat issue with IR & autorepeat

### DIFF
--- a/src/input/InputLIRC.cpp
+++ b/src/input/InputLIRC.cpp
@@ -107,7 +107,8 @@ void InputLIRC::read(int handle)
       // it makes GUI unusable
       if ((repeatCount % 3) == 0)
       {
-        emit receivedInput("LIRC", command);
+        bool up = (command.endsWith("_UP") && (command != "KEY_UP"));
+        emit receivedInput("LIRC", command, !up);
       }
     }
     else


plexappbuilder:
:balloon: - Good work!

src/input/InputComponent.h
```diff
@@ -84,6 +84,7 @@ class InputComponent : public ComponentBase
   const char* componentName() override { return "input"; }
   bool componentExport() override { return true; }
   bool componentInitialize() override;
+  bool supportsKeyUpDown() { return true; }
```
tru: you should mark this as `override` it's the C++11 version of virtual.

src/input/InputLIRC.h
```diff
@@ -22,6 +22,7 @@ class InputLIRC : public InputBase
 
   virtual const char* inputName() { return "LIRC"; }
   virtual bool initInput();
+  virtual bool supportsKeyUpDown() { return false; }
```
tru: All virtual here should be changed to `override` like everywhere else.

plexappbuilder:
:balloon: - Good work!

PR #286, opened 2016-03-30
Adapt InputSDL to new InputComponent

LongChair:
Allows to have again a proper autorepeat
The analog values handling has been removed, because it's not used.


plexappbuilder:
:balloon: - Good work!

PR #290, opened 2016-03-31
Don't assume cflags test for them instead.

tru:
Instead of maintaining lists per compiler of flags that should
be enabled / disabled we now test for them instead. If they are
able to be used we append them.

Same with linker flags we test for them before assuming they work.


plexappbuilder:
:-1: Tests failed!

plexappbuilder:
:balloon: - Good work!

plexappbuilder:
:balloon: - Good work!

PR #301, opened 2016-04-04
Map input to several actions

### LINKED ISSUE #214: Key input messages should carry multiple keys
Currently, only the first match in input mapping file triggers the event for a key press.
For keyboard input to work in search (and other possible text inputs) we need to change this so there would be multiple key names in the payload: one with a letter and one (if exists) with matching logical key.

Implementation proposal is to add a line like `"(?:Shift\\+)?([A-Z])": "%1",` to `keyboard.json` and change triggering code to not stop after first match but to collect every matched keys.
Desired API change: use existing endpoint, just change the argument from string to array of strings.

Testing: currently this model is supported in `tv-key-handling-refactoring-9001` branch of web repo. It's intended to be merged really soon.

tru:
This allows multiple actions for one button to be sent to the web-client. Necessary for getting search to work properly

Fixes #214

### DIFF
--- a/resources/inputmaps/keyboard.json
+++ b/resources/inputmaps/keyboard.json
@@ -17,6 +17,9 @@
     // map X and Shift+X to X. This allows normal number input
     "(?:Shift\\+)?([0-9])": "%1",
 
+    // map X and Shift+X to X for letters. This allows normal text input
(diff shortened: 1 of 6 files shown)


plexappbuilder:
:balloon: - Good work!

PR #302, opened 2016-04-05
Support PHT style harmony apple remote input

tru:
This following code emulates the harmony remotes input to PHT. The Apple API is actually
limited to just the number of keys that are on Apple specific remotes so we had to work
around that with the following hack: deviceID was used as a prefix for what the keys meant
so for example if we send deviceID 150 and keycode UP it means up, but instead deviceID 151
and UP it can be mapped to a different key.

The harmony profile for this is using deviceID's 150-160 to prefix the keycode, so we will
treat them differently here than the other keycodes. We append the deviceID to the keycode
when the deviceID falls inside the 150-160 range, otherwise we assume the keycode is coming
from a "normal" remote and only forward the keycode.

Since it's unknown if this will cause problems with any remotes I have added a setting:
appleremote.emulatepht to turn this off if needed, but for now we'll keep it defaulted to on


plexappbuilder:
:-1: Tests failed!

plexappbuilder:
:-1: Tests failed!

tru:
jenkins

plexappbuilder:
:-1: Tests failed!

tru:
jenkins

plexappbuilder:
:-1: Tests failed!

plexappbuilder:
:-1: Tests failed!

tru:
jenkins

plexappbuilder:
:-1: Tests failed!

CMakeModules/WebClientResources.cmake
```diff
@@ -22,7 +22,6 @@ if(NOT SKIP_WEB_CLIENT)
       EXPECTED_HASH SHA1=${WEB_CLIENT_HASH}
       TIMEOUT 100
       SHOW_PROGRESS
-      TLS_VERIFY ON
```
LongChair: Just curious, why did we remove that one ?

tru:
jenkins

plexappbuilder:
:balloon: - Good work!

Issue #283, opened 2016-03-30
Add shutdown/suspend/reboot to OSX and Windows


tru:
We only support these on OpenELEC right now. Many users want to do it on all platforms. It should be easy to do by extending the PowerComponent classes for OSX and Windows.

@LongChair have a look at this when you are done with all the input changes.

LongChair:
this is in master now, closing.

PR #304, opened 2016-04-06
Add shutdown/suspend/reboot to Windows, part of #283

LongChair:


plexappbuilder:
:-1: Tests failed!

LongChair:
Jenkins

plexappbuilder:
:balloon: - Good work!

src/power/PowerComponentWin.cpp
```diff
@@ -13,3 +15,52 @@ void PowerComponentWin::doEnableScreensaver()
...
+{
+  if (!WinUtils::getPowerManagementPrivileges())
+  {
+    QLOG_ERROR() << "Failed to obtain power management, priviledges";
+    return false;
+  }
+
+  if (QSysInfo::WindowsVersion >= QSysInfo::WV_WINDOWS8)
+  {
+    return InitiateShutdownW(NULL, NULL, 0, SHUTDOWN_INSTALL_UPDATES | SHUTDOWN_RESTART,
```
tru: This get very verbose maybe we can collect the common flags from all API calls in a constant above and just add the specific flag for each function when calling. Should make the calls a bit cleaner.

src/power/PowerComponentWin.cpp
```diff
@@ -13,3 +15,52 @@ void PowerComponentWin::doEnableScreensaver()
 {
   SetThreadExecutionState(ES_CONTINUOUS);
 }
+
+///////////////////////////////////////////////////////////////////////////////////////////////////
+bool PowerComponentWin::Suspend()
+{
+
```
tru: space?

src/power/PowerComponentWin.cpp
tru: Also if we can get the function call on one line you can drop the `{ }`

src/utils/win/WinUtils.cpp
```diff
@@ -0,0 +1,30 @@
+#include "WinUtils.h"
+#include "windows.h"
+
+/////////////////////////////////////////////////////////////////////////////////////////
+bool WinUtils::getPowerManagementPrivileges()
```
tru: Is this available for all users or will we have some UAC fuckage here?

LongChair: From what i have read, it should be, just that this priviledge is not a default one for processes.

src/power/PowerComponentWin.cpp
```diff
@@ -13,3 +15,52 @@ void PowerComponentWin::doEnableScreensaver()
...
+///////////////////////////////////////////////////////////////////////////////////////////////////
+bool PowerComponentWin::PowerOff()
+{
+  if (!WinUtils::getPowerManagementPrivileges())
+  {
+    QLOG_ERROR() << "Failed to obtain power management, priviledges";
+    return false;
+  }
+
+  if (QSysInfo::WindowsVersion >= QSysInfo::WV_WINDOWS8)
```
dogonwheels: Would it be shorter/less copy-paste-prone to set the flags based on platform, rather than the call?

LongChair: That os probably a good idea :)

src/power/PowerComponentWin.cpp
dogonwheels: I'm likely going daft - but I don't see the diff here. Aren't these the same calls/flags?

tru: the first one is different `SHUTDOWN_RESTART` `SHUTDOWN_HYBRID` etc.

tru: depending on the function of course. I was thinking that we could have to global for all functions and then just add the actual modifier.

LongChair: definitely a copy/paste issue here ... thanks @dogonwheels :p

plexappbuilder:
:balloon: - Good work!

src/power/PowerComponentWin.h
```diff
@@ -9,6 +10,14 @@ class PowerComponentWin : public PowerComponent
   PowerComponentWin() : PowerComponent(0) { }
   virtual void doDisableScreensaver();
   virtual void doEnableScreensaver();
+
+  virtual bool canPowerOff() override { return WinUtils::getPowerManagementPrivileges(); }
```
tru: Hmm, do we really want to call this every time we show the dialog? Wouldn't it be better to init this once and save the result so that we know if it was successful or not?

src/power/PowerComponentWin.cpp
```diff
@@ -13,3 +14,29 @@ void PowerComponentWin::doEnableScreensaver()
...
+///////////////////////////////////////////////////////////////////////////////////////////////////
+bool PowerComponentWin::PowerOff()
+{
+  DWORD shutdownFlags = SHUTDOWN_INSTALL_UPDATES | SHUTDOWN_POWEROFF;
+
+  if (QSysInfo::WindowsVersion >= QSysInfo::WV_WINDOWS8)
+    shutdownFlags |= SHUTDOWN_HYBRID;
+
+  return InitiateShutdownW(NULL, NULL, 0, shutdownFlags,
+                           SHTDN_REASON_MAJOR_APPLICATION | SHTDN_REASON_MINOR_OTHER | SHTDN_REASON_FLAG_PLANNED) == ERROR_SUCCESS;
```
tru: `SHTDN_REASON_MAJOR_APPLICATION | SHTDN_REASON_MINOR_OTHER | SHTDN_REASON_FLAG_PLANNED` can be put in a constant and used in both here and the place above.

src/power/PowerComponentWin.h
LongChair: Dunno if it's really heavy, but i can move that to the componentInitialize()

src/power/PowerComponentWin.cpp
LongChair: ok, makes sense

src/power/PowerComponentWin.h
tru: Yeah who knows with windows API. if we only need to call it once, let's just call it once.

tru:
Handled in the private staging.

PR #305, opened 2016-04-06
Expose sounds to the local http server

tru:
You can now request /sounds/$nameofsound that will first grab
sounds from the dataDir/sounds dir and secondly from the internal
resources.


plexappbuilder:
:balloon: - Good work!

PR #306, opened 2016-04-06
Fix Roku input so that keypresses works as expected

tru:


plexappbuilder:
:balloon: - Good work!

PR #339, opened 2016-05-16
Pass `-std=c++11` to build

jbergstroem:
`Apple LLVM version 7.3.0 (clang-703.0.16.1)` complained while referencing QT if `c++11` wasn't passed.

Not sure if this is where you want this, but I couldn't get it to build otherwise.


tru:
I think I fixed this on another branch already. The correct way is to the the CXX_STANDARD on the targets not passing the flag directly since it can differ between compilers. There is a add_std_properties() or something similar that's missing for one target plistplayer maybe.

jbergstroem:
I thought so much; at least I know its being worked on. Thanks.

PR #344, opened 2016-05-21
Adds an "Always on top" option to the video player

lukele:
- The status of "Always on Top" can be toggled via the Menu Bar

While I added this feature for the OS X version of Plex Media Player, based on the QT docs I believe this should work across platforms. I haven't had a chance to verify that, though.

I'd love to see this feature integrated into Plex Media Player and hope that some of you might find it useful. I've never coded in C++ before, so please don't judge the style to harshly.


tru:
Hi! Thanks for your effort! I think in general the functionallity is good and something we want to have in PMP. But I think it needs to be hooked up to our settings system so that the change can be accessed from a remote control as well. Also I think adding the menu bar will be strange on Linux and Windows since they are added inside the window. See how the full screen toggle is handled. 

There are some formatting issues with the code as well with tabs vs spaces etc but I can nitpick that later :)

tru:
Btw I am not against exposing settings in the menu bar but it should not be a ad-hoc solution rather added on top of our current settings component so you can flag stuff to be visible in the menu bar.

wm4:
This also lacks the CLA stub, I think: https://forums.plex.tv/discussion/188192/plex-media-player-open-source-contributions-and-building-faq

lukele:
Hi @tru,

I absolutely realize that the menu bar might no be the optimal solution. I decided against the settings panel in this first approach, because I think that "Stay on Top" is a very temporary solution rather than a fixed one, and having to go into the settings panel each time would be cumbersome (also, my C++ knowledge is very limited, and adding the menu bar option was very trivial).

Of course, by assigning a keyboard shortcut for the option in the settings panel, it would solve the steps necessary to quickly switch the setting.

Thanks for the input!

lukele:
@wm4: the link doesn't seem to be working for me. Could you share some info what exactly is missing?

tru:
Here is the link for the CLA: https://github.com/plexinc/plex-media-player/wiki/Contributor-License-Agreement

To add something to our settings screen you need to first add it here: https://github.com/plexinc/plex-media-player/blob/master/resources/settings/settings_description.json

Then you can look for how the fullscreen flag is handled in KonvergoWindow.cpp and you should do something similar there.

You can then add a keyboard shortcut by adding a action like we do with the zoom one: https://github.com/plexinc/plex-media-player/blob/master/resources/inputmaps/keyboard.json#L60

lukele:
Fantastic thanks!
I'll have a look at the documentation and will update my code accordingly (also trying to adjust the style)

lukele:
The updated code implements the always on top option now as a setting. In addition the shortcut CMD+Shift+A or T can be used to toggle the state.
The code style should also be adjusted to better match the Plex coding guidelines. 

What I couldn't find out is how to give the setting a human readable name (I assume that's not in the open source part?)

@tru You mentioned a solution to flag settings to be exposed via a menu bar. Is such a system already in place? If it is, could you elaborate on how to hook into it?

What is still missing is some kind of visual feedback that reflects the state update. I think it could be a simple overlay (similar to the loading spinner overlay). Could you give me some hints as how to add that best and where?

Last but not least, since the first commits were basically undone by the last one and also didn't include the CLA statement, what's the best way to proceed here?

lukele:
@tru 

Seeing that I made an ugly mess out of this pull request by not working on a separate branch and correctly integrating upstream from time to time, I was wondering if it makes the most sense if I close this one and open a new one

tru:
Yeah I was going to ask you to update it after our latest push. Please recreate it if you want.

tru:
I am likely to merge it after that so it won't be in vain :)

lukele:
Recreated this pull request at plexinc/plex-media-player#353 as one commit.
Closing this one.

PR #353, opened 2016-06-04
Add "Always on top" allowing PMP to float above any other windows

lukele:
- Allow the Plex Media Player window to flow above any other windows
- Expose a "alwaysOnTop" control in the settings panel to change the default state
- Allow the always-on-top state to be toggled using the shortcut CMD|Ctrl+Shift+A or T

Plex-CLA-1.0-signed-off-by: Lukas Pitschl <lukas@leftandleaving.com>

### DIFF
--- a/resources/inputmaps/keyboard.json
+++ b/resources/inputmaps/keyboard.json
@@ -52,14 +52,14 @@
     "Ctrl\\+Q": "host:close",
     "Ctrl\\+Shift\\+R": "host:reload",
     "Ctrl\\+Shift\\+D": "host:toggleDebug",
-    "Ctrl\\+Shift\\+A": "host:alwaysOnTop",
+    "Ctrl\\+Shift\\+A": "host:cycle_setting main.alwaysOnTop",
 
     // emulate some of of PHT's behaviour
     "I": "host:toggleDebug",
     "W": "toggle_watched",
     "\\\\": "host:fullscreen",
     "Z": "host:cycle_setting video.aspect",
-    "T": "host:alwaysOnTop",
+    "T": "host:cycle_setting main.alwaysOnTop",
 
     // media keys from the FLIRC and on Linux keyboards
     "Toggle Media Play\\/Pause": "play_pause",
--- a/resources/inputmaps/keyboard.json
+++ b/resources/inputmaps/keyboard.json
@@ -52,12 +52,14 @@
     "Ctrl\\+Q": "host:close",
     "Ctrl\\+Shift\\+R": "host:reload",
     "Ctrl\\+Shift\\+D": "host:toggleDebug",
+    "Ctrl\\+Shift\\+A": "host:alwaysOnTop",
 
     // emulate some of of PHT's behaviour
     "I": "host:toggleDebug",
     "W": "toggle_watched",
     "\\\\": "host:fullscreen",
     "Z": "host:cycle_setting video.aspect",
+    "T": "host:alwaysOnTop",
 
     // media keys from the FLIRC and on Linux keyboards
     "Toggle Media Play\\/Pause": "play_pause",
--- a/src/ui/KonvergoWindow.cpp
+++ b/src/ui/KonvergoWindow.cpp
@@ -272,7 +272,7 @@ void KonvergoWindow::updateAlwaysOnTopState()
 {
   QLOG_DEBUG() << "Changing always-on-top state";
   Qt::WindowFlags forceOnTopFlags = Qt::WindowStaysOnTopHint;
-#ifdef Q_OS_LINUX
+#ifdef Q_WS_X11
   forceOnTopFlags = forceOnTopFlags | Qt::X11BypassWindowManagerHint;
 #endif
 
--- a/src/ui/KonvergoWindow.cpp
+++ b/src/ui/KonvergoWindow.cpp
@@ -35,6 +35,7 @@ KonvergoWindow::KonvergoWindow(QWindow* parent) : QQuickWindow(parent), m_debugL
   InputComponent::Get().registerHostCommand("toggleDebug", this, "toggleDebug");
   InputComponent::Get().registerHostCommand("reload", this, "reloadWeb");
   InputComponent::Get().registerHostCommand("fullscreen", this, "toggleFullscreen");
+  InputComponent::Get().registerHostCommand("alwaysOnTop", this, "toggleAlwaysOnTop");
 
 #ifdef TARGET_RPI
   // On RPI, we use dispmanx layering - the video is on a layer below Konvergo,
@@ -70,6 +71,9 @@ KonvergoWindow::KonvergoWindow(QWindow* parent) : QQuickWindow(parent), m_debugL
 #else
   updateFullscreenState(false);
 #endif
+
+  // Check the always on top setting and activate it if necessary.
+  updateAlwaysOnTopState();
 
   emit enableVideoWindowSignal();
 }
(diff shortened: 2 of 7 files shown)


tru:
@lukele - Sorry for the silence. The 1.1.2 RC process took longer than expected. But now I will review it for you :-)

resources/inputmaps/keyboard.json
```diff
@@ -52,12 +52,14 @@
     "Ctrl\\+Q": "host:close",
     "Ctrl\\+Shift\\+R": "host:reload",
     "Ctrl\\+Shift\\+D": "host:toggleDebug",
+    "Ctrl\\+Shift\\+A": "host:alwaysOnTop",
```
tru: I think you should be able to use `host:cycle_setting alwaysOnTop` here that way you won't have to add a new host command.

src/ui/KonvergoWindow.cpp
```diff
@@ -227,6 +240,11 @@ void KonvergoWindow::updateMainSectionSettings(const QVariantMap& values)
     SystemComponent::Get().setCursorVisibility(!SettingsComponent::Get().value(SETTINGS_SECTION_MAIN, "disablemouse").toBool());
   }
 
+  if (values.find("alwaysOnTop") != values.end())
+  {
```
tru: you can remove brackets here when it's only one statement in the if clause.

src/ui/KonvergoWindow.cpp
```diff
@@ -253,6 +271,27 @@ void KonvergoWindow::updateFullscreenState(bool saveGeo)
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
+void KonvergoWindow::updateAlwaysOnTopState()
+{
+  QLOG_DEBUG() << "Changing always-on-top state";
+  Qt::WindowFlags forceOnTopFlags = Qt::WindowStaysOnTopHint;
+#ifdef Q_OS_LINUX
+  forceOnTopFlags = forceOnTopFlags | Qt::X11BypassWindowManagerHint;
```
tru: This is a X11 specific thing right? In that case you should probably use Q_WS_X11 otherwise it happens on raspberry pi as well.

src/ui/KonvergoWindow.cpp
```diff
@@ -253,6 +271,27 @@ void KonvergoWindow::updateFullscreenState(bool saveGeo)
...
+#endif
+
+  if(SettingsComponent::Get().value(SETTINGS_SECTION_MAIN, "alwaysOnTop").toBool())
+  {
```
tru: remove brackets here.

tru:
Some minor issues but all in all looks really good. Thanks for looking into this!

lukele:
Thank you! I'll integrate your proposed changes on the weekend and will update the pull request.

tru:
Hey @lukele have you had time to look at this again? I am keen on getting it in.

lukele:
Hi @tru 

I've finally come around to making the changes.
Always on top is now implemented as cycleable setting.
While I do like the reduction in code, I believe the previous version might have been better from a user experience point. While alwaysOnTop was previously displayed as a checkbox in the settings pane, it's now displayed as a select menu with options "Enabled" or "Disabled".

Also I couldn't figure out how to add proper strings to this small dialog that pops up whenever changing the always on top status:

plex overlay

tru:
That string there probably just need a translation in the web-client, strange it shows up as undefined instead of the key though, we should check with the web guys about that.

I think that an optimal solution would be to have cycle_setting handle booleans as well. This should be a pretty easy change in `SettingsComponent::cycleSetting`. If you are up for it feel free to make that change, otherwise I can accept this PR as is and fix it myself later.

Thanks for you work on it. It looks good now overall!

lukele:
I've modified cycleSetting to accept booleans now as well.
The modified implementation will assume that if no possible values are defined, the setting is a boolean. I'm not sure whether that's an over simplified interpretation or not.

AlwaysOnTop is now back to being displayed as a checkbox in Settings.

tru:
The way we do it in other places is to check the type of the `default` value. It can be string or boolean.

And the implementation looks exactly what I imagined :)

lukele:
Great to hear :)

I've tried to figure out how to access the default value from SettingsComponent.
From what I could tell, it would be possible via the SettingsSection->m_values. That variable is however protected, so it's not currently possible to access it from SettingsComponent.

Could you give me a hint on how to get to the default value and check for its type in SettingsComponent?

tru:
Hmm, I see now that we don't expose that anymore for whatever reason. I think the solution is to add `defaultValue(QString)` in `SettingsSection` that calls `m_values[key]->defaultValue()` that gives you the default and then you can check the type on that one (it's a QVariant).

Something like this diff: https://gist.github.com/bd55935d43abb039dcac81d25b17ebc8

lukele:
The explicit check of the defaultValue is now added.
As you'll see, I'm typecasting the return of QVariant.type() to QMetaType::Type. According to the Qt docs, that should be the way to go.
I've also found however, that in this case, the cast would not be necessary if the comparison uses QVariant::Bool instead of QMetaType::Bool.

Which version do you prefer? Both seem to work flawlessly.

tru:
I think we have so far been using comparison against QVariant::Bool and not casted it to QMetaType::Type. But I don't think it matters that much. I would be happy to merge this as is if you feel you are done.

lukele:
I've removed the type cast. It looks cleaner like this :)
Besides the translation of the strings which have to be done in the web-client I believe, I think it's done.
Thank you for all your input and pointers

tru:
Merged! Thanks for all your input.

xanderiel:
Sorry for the stupid question, but is this in the main build and where exactly do I find it?

wm4:
It is. You can download it from plex.tv, build it yourself, or use one of the 3rd party builds that are floating around.

xanderiel:
Excuse my asking, but I downloaded the Plex Media Server for Mac OS X on https://www.plex.tv/downloads/ and in the Settings I cannot find it.

1) Download

image

2) Settings

image

wm4:
It's available only in TV mode.

xanderiel:
For anyone trying this out:
1) Start the Plex Media Player app on Mac OS X.
2) Go into *Fullscreen TV Mode* by pressing `Cmd+Shift+F`.
3) Go to `Profile Icon » Settings » Main » Always on Top: Check`

As an interim step, you'll now have Plex' "TV mode" which takes up its own Desktop Space in Mac OS X anyway, so this does not solve the "Always on Top" thing when you want to browse the internet and also watch a video. But there's a hack of sorts. In case you would now press `Cmd+Shift+F` again, you would leave the precious TV mode which allows the Always on Top feature and go back to the standard Plex app which just seems to be a WebView of the Web Client anyway. But you can tell Mac OS X to leave the fullscreen mode and enjoy both Plex' TV mode with Always on Top while doing something else on any desktop. Feel free to combine with `Assign to All Desktops` to have it on every desktop on top of everything else.

4) Press `Cmd+ctrl+F` (not Plex' `Cmd+Shift+F`, but Mac OS X' `Cmd+ctrl+F`)
5) Optionally, right-click the Plex icon in the Mac OS Dock, then `Assign to » All Desktops`.

There is one issue though, don't minimize the window (in version 3.5.0 or lower presumably) since it will not come back up. You need to restart the app.

lukele:
To add some more information. It's possible to have TV Mode in non-fullscreen if you change the setting ``Profile Icon » Settings » Main » Layout`` to ``TV``
You can then also freely toggle between "Always on Top" *Off* and *On* with ``CMD+Shift+A``

drewwells:
Thanks for the workaround. This seems something very useful for the web view too though, as that is the case where always on top significantly changes the usefulness

PR #400, opened 2016-08-12
Allow to enable D-Bus power management using a CMake option

jleroy:
At this time, only OpenELEC builds can use D-Bus for power-related operations (Shutdown/Reboot/Suspend).

This pull request allow to enable D-Bus on any Linux distribution using a new `-DLINUX_DBUS` option at compile time.

For more clarity, I've renamed the `PowerComponentOE`class to `PowerComponentDBus`. The `LINUX_DBUS` option is always enabled for OpenELEC builds.


wm4:
Isn't that missing some code?

jleroy:
Everything is here. Why?

wm4:
I misunderstood the intention of this PR. I think this should be ok then.

wm4:
Merged, thanks.

PR #405, opened 2016-08-21
Add AML as a possible target

LongChair:
This allows to add AML changes so that pmp can be run on these platforms.


LongChair:
outdated.

blkqi:
Is there any chance that this effort can be revived?

PR #426, opened 2016-10-06
Fix slow RPi playback

jabbera:
Fix for: https://forums.plex.tv/discussion/237626/plex-media-player-1-1-5-1-1-6-is-now-released/p2

Was removed in: bb53026


wm4:
Our plan is to use the new code. It was fine in our tests, but for some reason not fine upon release. We're having a closer look.

PR #484, opened 2016-12-20
Updated macOS build instructions to include conan

mseeley:
Previously the `cmake` command would fail with a conan error. Following the Linux instructions to install conan would fail in `pip` with an error. Using `brew` to install conan works well on 10.12


wm4:
LGTM

PR #511, opened 2017-02-02
Fix CGDisplayMode comparison on macOS

FrostedMint:
This fixes an issue I encountered on macOS where the current display mode could not be determined, causing the Display FPS to be set to 0, and preventing Display Sync (and mpv’s interpolation feature) from working correctly.

Here’s the associated error I was seeing in my logs:
`DisplayManager : unable to retrieve current video mode`

This change attempts to compare CGDisplayModes using the IOKit ID, rather than comparing the CGDisplayMode references.

I tested this fix by switching between displays and switching resolutions; the current mode identified on the debug screen appears to update correctly, mpv’s interpolation appears to work correctly, and the log error is no longer present. My system is running 10.12.3 on 2016 MBP 13”.

This is my first time working with macOS APIs, so apologies if anything is isn’t quite right.

### DIFF
--- a/src/display/osx/DisplayManagerOSX.cpp
+++ b/src/display/osx/DisplayManagerOSX.cpp
@@ -120,10 +120,14 @@ int DisplayManagerOSX::getCurrentDisplayMode(int display)
     return -1;
   
   CGDisplayModeRef currentMode = CGDisplayCopyDisplayMode(m_osxDisplays[display]);
+  uint32_t currentIOKitID = CGDisplayModeGetIODisplayModeID(currentMode);
 
   for (int mode = 0; mode < CFArrayGetCount(m_osxDisplayModes[display]); mode++)
   {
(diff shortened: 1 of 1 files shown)


wm4:
So I suppose mode references don't have to be the same. Pretty nice catch. Pinging @LongChair for review.

Also, can you sign off with the CLA? See: https://github.com/plexinc/plex-media-player/wiki/Contributor-License-Agreement

LongChair:
LGTM, if those IOKit IDs are unique for each mode, that should be fine :)

FrostedMint:
I’m not comfortable publicly associating my legal name with this account. Is there any way I could sign off on the CLA privately/out-of-band?

Re IOKit IDs: they appear to be unique, but the Apple docs are frustratingly vague.

tru:
@FrostedMint sure - just add the header with you signing off with your alias and email tobias at plex tv referencing it and acking your real name to me. I will just make a note of it.

FrostedMint:
@tru Done, thanks!

wm4:
I tested this, and it seems to work well. Merged, thanks!

PR #586, opened 2017-07-24
KonvergoWindow: allow the default handling of focus events to occur

rcombs:
This allows Chromium to be notified when the window is focused and blurred,
which fixes some awkward issues where it'd keep modifier keys "held" after
you tabbed away from PMP and back.


wm4:
OK. Seems more like an oversight than intentional.

PR #587, opened 2017-07-24
Energy consumption improvements

rcombs:
This is an assortment of changes meant to reduce the background CPU utilization of PMP.

### DIFF
--- a/src/input/InputRoku.cpp
+++ b/src/input/InputRoku.cpp
@@ -66,11 +66,11 @@ void InputRoku::ssdpRead()
 void InputRoku::parseSSDPData(const QByteArray& data, const QHostAddress& sender, quint16 port)
 {
   if (data.contains("M-SEARCH * HTTP/1.1"))
-    m_ssdpSocket->writeDatagram(getSSDPPacket(), sender, port);
+    m_ssdpSocket->writeDatagram(getSSDPPacket(sender), sender, port);
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
-QByteArray InputRoku::getSSDPPacket()
+QByteArray InputRoku::getSSDPPacket(const QHostAddress& sender)
 {
   QByteArray packetData;
 
@@ -82,7 +82,7 @@ QByteArray InputRoku::getSSDPPacket()
   packetData.append("Server: Roku UPnP/1.0 MiniUPnPd/1.4\r\n");
 
   packetData.append("Location: http://");
-  packetData.append(Utils::PrimaryIPv4Address());
+  packetData.append(sender.toString());
   packetData.append(":8060/\r\n");
 
   packetData.append("USN: uuid:roku:ecp:");
--- a/src/utils/Utils.cpp
+++ b/src/utils/Utils.cpp
@@ -13,6 +13,8 @@
 #include <QFile>
 #include <QSaveFile>
 
+#include <mutex>
+
 #include "settings/SettingsComponent.h"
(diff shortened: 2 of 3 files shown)


src/input/InputSDL.cpp
```diff
@@ -182,13 +191,41 @@ void InputSDLWorker::run()
...
+    m_mutex.unlock();
   }
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
+void InputSDLWorker::wake()
+{
+  m_mutex.lock();
+  m_waitCond.wakeAll();
+  m_mutex.unlock();
```
wm4: Somewhat questionable use of condition variables. This is racy. It will wake up threads that are currently waiting, but not threads that are about to enter a wait call. It might actually be fine in this case, though - if you think it is, fine.

src/input/InputRoku.cpp
```diff
@@ -82,7 +82,7 @@ QByteArray InputRoku::getSSDPPacket()
   packetData.append("Server: Roku UPnP/1.0 MiniUPnPd/1.4\r\n");
 
   packetData.append("Location: http://");
-  packetData.append(Utils::PrimaryIPv4Address());
+  packetData.append(sender.toString());
```
wm4: Unfamiliar with this code, but the commit looks good to me.

src/utils/Utils.cpp
```diff
@@ -39,11 +41,16 @@ QString Utils::sanitizeForHttpSeparators(const QString& input)
...
 #ifdef Q_OS_MAC
-  return OSXUtils::ComputerName();
+    name = OSXUtils::ComputerName();
 #else
-  return QHostInfo::localHostName();
+    name = QHostInfo::localHostName();
 #endif
+  });
+  return name;
 }
```
wm4: Is there any kind of scenario where the computer name changes and we don't want to require a PMP restart? I guess not?

rcombs: I figure it's an a rare enough case that we can just say you need to restart it, yeah.

src/input/InputSDL.cpp
rcombs: I'm locking around the `QGuiApplication::focusWindow()` check, so this should be fine assuming that returns true before/while `QQuickWindow::focusInEvent` is called, which I should verify.

wm4: Hm, ok.

Another thing. Looking at this again, ``QGuiApplication::focusWindow()`` seems to be called from another thread. Is the method thread-safe at all? Would it be safer to manage this in the main thread instead, maybe with a proper condition?

LongChair:
This is probably out of topic, but we should probably check QtGamepad now that it exists and check how well it would work with usual controllers ;)

rcombs:
Hmm, it looks like that actually does use async APIs on macOS, Windows, and Linux where available (rather than polling), so it'd definitely be a better option.

src/input/InputSDL.cpp
rcombs: Turns out, nope, it's not, which seems somewhat silly. I'm a little out of my depth on the Qt stuff; can you take this?

wm4: Do you have a way to test the SDL code? (I don't.)

I could make a test branch that tries to notify the SDL thread of focus changes.

HazCod:
Any plans for merging this?

mseeley:
These commits landed as part of #778

PR #588, opened 2017-07-24
Initial support for Bluetooth and Touch Bar controls on macOS

rcombs:
This uses a private API to keep the Touch Bar controls active in the foreground
Both iTunes and Safari do the same; Safari source here:
https://github.com/WebKit/webkit/blob/bde8ac5bddc49527396f3adb576d9a101a8e4828/Source/WebCore/platform/audio/mac/MediaSessionManagerMac.mm#L202


src/player/PlayerComponent.cpp
```diff
@@ -1084,6 +1085,9 @@ void PlayerComponent::updateVideoAspectSettings()
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 void PlayerComponent::updateVideoSettings()
 {
+  if (!m_mpv)
+    return;
```
wm4: (Huh?)

src/input/apple/InputAppleMediaKeys.mm
```diff
@@ -69,9 +122,97 @@ -(void)mediaKeyTap:(SPMediaKeyTap *)keyTap receivedMediaKeyEvent:(NSEvent *)even
...
+    void* lib = dlopen("/System/Library/PrivateFrameworks/MediaRemote.framework/MediaRemote", RTLD_NOW);
+    if (lib) {
+#define LOAD_FUNC(name) \
+  name = (name ## Func)dlsym(lib, "MRMediaRemote" #name)
+      LOAD_FUNC(SetNowPlayingVisibility);
+      LOAD_FUNC(GetLocalOrigin);
+      LOAD_FUNC(SetCanBeNowPlayingApplication);
+      if (SetCanBeNowPlayingApplication)
+        SetCanBeNowPlayingApplication(1);
+    }
```
wm4: I suppose these are all private API functions. This looks OK - if Apple leaves us no choice, it's probably just Apple being Apple, and we can gracefully handle the unlikely event of Apple removing or hiding these functions.

resources/inputmaps/apple-media-keys.json
```diff
@@ -3,7 +3,9 @@
   "idmatcher": "AppleMediaKeys.*",
   "mapping":
   {
-    "KEY_PLAY": "play_pause",
+    "KEY_PLAY": "play",
+    "KEY_PAUSE": "pause",
+    "KEY_PLAY_PAUSE": "play_pause",
```
wm4: Slightly worried about unintended side-effects. Did remotes using this config file always emit KEY_PLAY_PAUSE, and did these keys just match KEY_PLAY before?

src/player/PlayerComponent.cpp
rcombs: I ran into a crash in here while testing!

wm4: Feels like this could happen only during construction or destruction at best.

resources/inputmaps/apple-media-keys.json
rcombs: Hmm, I think the only user is this: https://github.com/plexinc/plex-media-player/blob/47e2f6e39cb8f76fb7a71cb230fbc56fc6627af9/src/input/apple/InputAppleMediaKeys.mm#L48
so we should be able to change that to `INPUT_PLAY_PAUSE` and be fine.

wm4: I mean, I'm worried about existing user input mappings. Will they break with this change? It sounds like they possibly could. Then again, I'm not too worried.

rcombs: Oh, I guess? Do people really override the AppleMediaKeys map?

wm4: Probably not, which is why I'm not overly worried.

Maybe it could be an entry in the changelog.

PR #600, opened 2017-08-22
PlayerComponent: resolve plex.direct domains internally

rcombs:
This also enables TLS certificate verification in mpv, like the rest of the app

### DIFF
--- a/src/player/PlayerComponent.cpp
+++ b/src/player/PlayerComponent.cpp
@@ -20,6 +20,7 @@
 #include <math.h>
 #include <string.h>
 #include <shared/Paths.h>
+#include <unistd.h>
 
 #ifdef TARGET_RPI
 #include <bcm_host.h>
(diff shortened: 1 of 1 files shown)


src/player/PlayerComponent.cpp
```diff
@@ -282,6 +297,13 @@ void PlayerComponent::queueMedia(const QString& url, const QVariantMap& options,
   extraArgs.insert("ad", "");
   extraArgs.insert("vd", "");
 
+  extraArgs.insert("tls-verify", "yes");
+  extraArgs.insert("stream-lavf-o", "verifyhost=" + host);
```
wm4: These don't really need to be per-media. They can be set as global options. (The actual difference isn't huge, but still.)

src/player/PlayerComponent.cpp
```diff
@@ -282,6 +297,13 @@ void PlayerComponent::queueMedia(const QString& url, const QVariantMap& options,
...
+
+#ifdef Q_OS_LINUX
+  extraArgs.insert("tls-ca-file", "/etc/ssl/cert.pem");
+#endif
```
wm4: I'm on Linux, but I don't have this file. Could be made OPENELEC only, or dynamically by checking whether the file exists.

src/player/PlayerComponent.cpp
```diff
@@ -243,6 +243,17 @@ bool PlayerComponent::load(const QString& url, const QVariantMap& options, const
...
 ///////////////////////////////////////////////////////////////////////////////////////////////////
+static QString ConvertPlexDirectURL(const QString& host)
+{
+    if (!host.endsWith(".plex.direct"))
+        return host;
+
+    QString substr = host.left(host.indexOf('.'));
+    substr.replace('-', '.');
+    return substr;
+}
```
wm4: Idea seems ok, but misses doing this for external subtitles.

src/player/PlayerComponent.cpp
rcombs: Ah, I looked up what Qt does and it searches a series of dirs; we could do something similar (and could do it once globally at startup).

src/player/PlayerComponent.cpp
rcombs: `tls-verify` and `tls-ca-file` can be; `verifyhost`, probably not, since it could differ between queue items?

src/player/PlayerComponent.cpp
rcombs: Ah! Re-pushed with a (untested) fix.

wm4:
PlayerComponent.cpp actually needs a `#include <unistd.h>`, otherwise LGTM.

PR #611, opened 2017-09-13
Fix typo in README.md

michael131468:
Change a reference of SLD to SDL.


wm4:
Thanks for the fix. Can you add the stuff in https://github.com/plexinc/plex-media-player/wiki/Contributor-License-Agreement to your commit message?

PR #666, opened 2018-02-21
Bump Qt (plus bonus energy consumption fixes)

rcombs:
Qt bump fixes relay.

### DIFF
--- a/CMakeModules/QtConfiguration.cmake
+++ b/CMakeModules/QtConfiguration.cmake
@@ -10,7 +10,7 @@ if(NOT IS_DIRECTORY ${QTROOT})
 		"plexmediaplayer-qt"
 		DIRECTORY dir
 		DEPHASH_VAR QT_DEPS_HASH
-    BUILD_NUMBER 165
+    BUILD_NUMBER 187
     ARTIFACTNAME konvergo-qt
     DYLIB_SCRIPT_PATH ${PROJECT_SOURCE_DIR}/scripts/fix-install-names.py
 	)
--- a/src/display/DisplayManager.cpp
+++ b/src/display/DisplayManager.cpp
@@ -79,8 +79,8 @@ bool DisplayManager::isValidDisplayMode(int display, int mode)
 // multiple: display FPS
 bool DisplayManager::isRateMultipleOf(float refresh, float multiple, bool exact)
 {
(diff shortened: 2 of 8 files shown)


LongChair:
Globally looks good to me.
I just have one question about the Qt bump. I am fine bumping it, I also have the equivalent embedded Qt bump over here :https://github.com/LongChair/LibreELEC.tv/commit/e5a628608cf8e44f0b9f9ec120d41ed89c6ecd8a

@markus101 : is that fine with you ? 

@rcombs : I would also need the plex.direct Qt patch to apply it on the embedded Qt build

rcombs:
It's in the deps builder repo. I think I have to add it to this repo as well.

LongChair:
hmmm is that those two hunks ? https://github.com/plexinc/plex-dependency-builder/blob/konvergo/patches/qt5-5.10.0.patch#L215-L261 . because that patch is a big bag :)

rcombs:
Yup, that's the ones. Also make sure the old patch isn't still included.

LongChair:
OK i'll check this shortly for embedded.
if @markus101 is fine with bumping Qt then go for it.
When you do it, Maybe just then push those commits to master without the merge, we always tried to avoid the merge commits :)

comio:
Very interesting patch. Can you wait a bit in order to give time to distros to upgrade to QT5.10? (I'm speaking as Gentoo linux user).

LongChair:
This doesn't break compatibility with Qt 5.9 anyways

comio:
Great.

markus101:
👍 from me, we'll get QA to take a pass at it before the next release.

der-eismann:
What's the status here? Does your QA take 3 months? :)

LongChair:
Qt was bumped to 5.9.5 in master.

HazCod:
@pcalves ? This PR did not only include a Qt bump.

pcalves:
I believe this branch was closed as a side effect of the `tv2` branch being deleted, definitely not intentional. 😞

I've changed the base branch to `master` and reopened the PR.

mseeley:
@rcombs let's tackle this again in https://github.com/plexinc/plex-media-player/issues/802. We'll need to bump all the way to at least 5.12 to account for Mojave compat.

PR #686, opened 2018-04-11
Install desktop files&icons

### LINKED ISSUE #625: Install Icons/Desktop File
It would be nice if Plexmediaplayer installed a .desktop file and the application icons itself on Linux

Cogitri:
Proposed patch to fix #625 . I'm not particularly fluent with CMake, so I'm sorry if this is incorrect - however it does work fine on my machine.

The .desktop file is from: https://github.com/Enigma0/build-scripts/blob/master/plexmediaplayer.desktop

### DIFF
--- a/CMakeModules/InstallLinuxDesktopFile.cmake
+++ b/CMakeModules/InstallLinuxDesktopFile.cmake
@@ -1,0 +1,2 @@
+install (FILES ${CMAKE_SOURCE_DIR}/resources/desktop/plexmediaplayer.desktop DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications)
+install (FILES ${CMAKE_SOURCE_DIR}/resources/images/icon.svg DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/icons/hicolor/scalable/apps RENAME plexmediaplayer.svg)
--- a/CMakeModules/LinuxConfiguration.cmake
+++ b/CMakeModules/LinuxConfiguration.cmake
@@ -18,8 +18,8 @@ else()
   Message(STATUS "Enabling D-Bus power management")
 endif()
 
-set(INSTALL_BIN_DIR bin)
-set(INSTALL_RESOURCE_DIR share/plexmediaplayer)
+set(INSTALL_BIN_DIR ${CMAKE_INSTALL_BINDIR})
+set(INSTALL_RESOURCE_DIR ${CMAKE_INSTALL_DATADIR}/plexmediaplayer)
 
 if(NOT OPENELEC)
   include(InstallLinuxDesktopFile)
--- a/CMakeModules/LinuxConfiguration.cmake
+++ b/CMakeModules/LinuxConfiguration.cmake
@@ -1,3 +1,5 @@
+include(GNUInstallDirs)
+
 find_package(X11)
 if(X11_FOUND AND X11_Xrandr_FOUND)
   include_directories(X11_X11_INCLUDE_PATH X11_Xrandr_INCLUDE_PATH)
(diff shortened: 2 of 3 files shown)


LongChair:
Hey @Cogitri , and thanks for your contribution

Will you please you please read through our CLA the update your commit message?For now the information is available on our wiki. https://github.com/plexinc/plex-media-player/wiki/Contributor-License-Agreement#individual-agreement

Cogitri:
Sure, I've added the legal notice for it. Just as a suggestion: Maybe it would be a good idea to add the content of that wiki page as contributing.md (or similar) to this repo so that GitHub prompts new contributors that such guidelines exist. See https://blog.github.com/2012-09-17-contributing-guidelines/ for more info.

LongChair:
Yeah that is a good suggestion :)

LongChair:
Another remark though, please make sure that your commit message described the context and what it’s fixing. Always better when we have to browse history :)

Cogitri:
Hm, I guess I'm just not creative with commit messages, but to me the commit message contains everything required. Maybe you could suggest an improvement? :)

LongChair:
It could at least mention what platform it’s about, what those files are for and what they do (ie what those files fix)
There is no need for being creative, just think as if you were in another context and read this trying to figure out what it does and what it’s about :)

CMakeLists.txt
```diff
@@ -69,6 +69,7 @@ elseif(WIN32)
   include(Win32Configuration)
 elseif(UNIX AND (NOT APPLE))
```
LongChair: We'd probably require a NOT OPENELEC test here as well as the desktop shortcut will be useless there.

CMakeLists.txt
```diff
@@ -69,6 +69,7 @@ elseif(WIN32)
...
   include(LinuxConfiguration)
+  include(InstallDesktopFiles)
```
LongChair: For clarity it would probably bettter that this gets into the LinixConfiguration cmake file like for other platforms.

CMakeModules/InstallDesktopFiles.cmake
```diff
@@ -0,0 +1,4 @@
+include(GNUInstallDirs)
+
+install (FILES ${CMAKE_SOURCE_DIR}/resources/desktop/plexmediaplayer.desktop DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications)
```
LongChair: I believe that this file sould be InstallLinuxDesktopFile which helps knowing it's about linux at first look

CMakeLists.txt
Cogitri: Hm, but than we'd have to have an `if(NOT OPENELEC)` for installing the desktop files, which would be rather confusing in the LinuxConfiguration file (since OpenELEC is Linux as well)

CMakeModules/InstallDesktopFiles.cmake
Cogitri: Yep, that sounds like a good idea

CMakeLists.txt
```diff
@@ -69,6 +69,9 @@ elseif(WIN32)
   include(Win32Configuration)
 elseif(UNIX AND (NOT APPLE))
   include(LinuxConfiguration)
+  if(NOT OPENELEC)
+    include(InstallLinuxDesktopFile)
+  endif(NOT OPENELEC)
```
LongChair: Sorry if i was unclear about this :)
But the `InstallLinuxDesktopFile.cmake` should be included from `LinuxConfiguration.cmake` as it's part of linux.
There is not problem making the OPENELEC check in `LinuxConfiguration.cmake` as we already make other checks for Linux related stuff ie X11 there.
Embedded is indeed Linux, but this would not appy for embedded (although adding shorcut wouldn't harm, but would be useless).

CMakeLists.txt
LongChair: this is not needed (seem comment below)

CMakeLists.txt
Cogitri: Ah, okay :)

PR #690, opened 2018-04-17
player: add playback speed control interface

LongChair:
This commit adds player playback speed control interface for web-client.

with `setPlaybackSpeed(double)` and `double playbackSpeed()`

Speed is defined as a speed factor ie 1.0 = normal playback speed, 2.0 is double speed etc …

### DIFF
--- a/src/player/PlayerComponent.cpp
+++ b/src/player/PlayerComponent.cpp
@@ -1524,3 +1524,17 @@ QString PlayerComponent::videoInformation() const
   return infoStr;
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////
+void PlayerComponent::setPlaybackSpeed(double speed)
+{
+  mpv::qt::set_property(m_mpv, "speed", speed);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+double PlayerComponent::playbackSpeed()
+{
+  QVariant speed = mpv::qt::get_property(m_mpv, "speed");
+  if (speed.isValid())
+    return speed.toDouble();
+  return 0;
+}
--- a/src/player/PlayerComponent.h
+++ b/src/player/PlayerComponent.h
@@ -119,6 +119,10 @@ public:
 
   QRect videoRectangle() { return m_videoRectangle; }
 
+  // Set the playback speed (1.0 = normal speed, 2.0 = 2x)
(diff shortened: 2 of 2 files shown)


src/player/PlayerComponent.cpp
```diff
@@ -1524,3 +1524,17 @@ QString PlayerComponent::videoInformation() const
...
+  mpv::qt::set_property(m_mpv, "speed", speed);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+double PlayerComponent::playbackSpeed()
+{
+  QVariant speed = mpv::qt::get_property(m_mpv, "speed");
+  if (speed.isValid())
+    return speed.toDouble();
+  return 0;
```
mseeley: C++ rookie question, the `speed` value set above is a `double`. But here the speed value returned in being handled as a `QVariant`. Is the treating `speed` as a `QVariant` just sugar to more easily ensure it's a non-null value?

What do you think of returning `1.0` when the speed is unknown? I'm concerned `0` is easy to conflate with a valid value. Additionally, if the UI has never called `setPlaybackSpeed` it seems reasonable that playback speed is a non-zero value (`1.0` being "normal").

src/player/PlayerComponent.cpp
```diff
@@ -1524,3 +1524,17 @@ QString PlayerComponent::videoInformation() const
   return infoStr;
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////
+void PlayerComponent::setPlaybackSpeed(double speed)
+{
+  mpv::qt::set_property(m_mpv, "speed", speed);
```
mseeley: Should this clamp to avoid negative double values?

sixones: MPV will most likely clamp the value.

src/player/PlayerComponent.cpp
sixones: > C++ rookie question, the speed value set above is a double. But here the speed value returned in being handled as a QVariant. Is the treating speed as a QVariant just sugar to more easily ensure it's a non-null value?

That's right, `QVariant` is a type that holds another type, without enforcing a type at compile time and can be passed around without forcing a specific type.

mseeley:
I'll adopt this one and land in January.

mseeley:
Closing due to inactivity.

PR #691, opened 2018-04-17
powercomponent: only report relaunch capability on embedded, fixes #354

### LINKED ISSUE #354: No "Exit" option in UI - only "Relaunch" and "Cancel" [Ubuntu 16.10 & SteamOS/Debian 8]
<strike>Not sure how wide this issue may be so I included Debian 8 as a reference.</strike> Issue effects Ubuntu/Kubuntu 16.04 & 16.10 as well.

When attempting to close the app, I am only met with two options: "Relaunch" and "Cancel"

Further, "Relaunch" doesn't actually ever relaunch the application.

**To be clear, there is an "Exit" option in the top right menu but the popup dialog where you actually make your choice has no such option.**

LongChair:
Reporting the relaunch capability gets web to display a relaunch menu entry.

That menu entry is only valid on Embedded where systemd will keep on relaunching upon PMP exit.
On linux builds, this will get reported as supported, and will therefore show the relaunch menu item.
When users click on it, it will simply exit, without relaunching.

We just make it valid for embedded with that patch.

cc @sixones

### DIFF
--- a/src/power/PowerComponent.h
+++ b/src/power/PowerComponent.h
@@ -34,9 +34,16 @@ public Q_SLOTS:
   bool canPowerOff() { return checkCap(CAP_POWER_OFF); }
   bool canReboot() { return checkCap(CAP_REBOOT); }
   bool canSuspend() { return checkCap(CAP_SUSPEND); }
-  bool canRelaunch() { return checkCap(CAP_RELAUNCH); }
+  bool canRelaunch()
+  {
+#if OPENELEC
+      return true;
+#else
+      return false;
+#endif
+  }
 
-  virtual int getPowerCapabilities() { return CAP_RELAUNCH; }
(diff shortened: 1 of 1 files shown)


src/power/PowerComponent.h
```diff
@@ -34,9 +34,15 @@ public Q_SLOTS:
   bool canPowerOff() { return checkCap(CAP_POWER_OFF); }
   bool canReboot() { return checkCap(CAP_REBOOT); }
   bool canSuspend() { return checkCap(CAP_SUSPEND); }
-  bool canRelaunch() { return checkCap(CAP_RELAUNCH); }
-
-  virtual int getPowerCapabilities() { return CAP_RELAUNCH; }
+  bool canRelaunch() {
```
sixones: `{` on the next line.

PR #693, opened 2018-04-18
icon: Do not enforce icon on startup, fixes #622

### LINKED ISSUE #622: PMP MacOS - icon is hardcoded, cannot change
The icon is too large compared to other items in my dock, so I've made a smaller Plex.icns.

Have put my custom Plex.icns in the PMP Finder app info, and have also replaced the Plex.icns in the app bundle /Contents/Resources

The new icon is fine while app is NOT running. But when PMP is running, it replaces my icon with its own in the Dock. I'm guessing the executable has a hardcoded icon that overrides my icon settings.

Would be great if the app could respect the icon settings the user has made. Thanks!

LongChair:
We enforce the icon onstartup and this has been reported as an issue for people willing to change the application Icon.

it seems that the application icon is anyways handled with the svg icon we have which is bundled into the app.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -165,7 +165,11 @@ int main(int argc, char *argv[])
       qputenv("QT_SCALE_FACTOR", scale.toUtf8());
 
     QApplication app(newArgc, newArgv);
+#if Q_OS_WIN
+    // Setting the icon on Windows is necessary but will break user
+    // ability to change icon on OSX
     app.setWindowIcon(QIcon(":/images/icon.png"));
+#endif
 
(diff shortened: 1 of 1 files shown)


LongChair:
cc @sixones

sixones:
Icon seems to be missing on the Windows now in the taskbar and on the titlebar on Windows 10.

image

LongChair:
@sixones : I limited setting the icon to windows platform so that it still fixes the OSX problem.

sixones:
Works well and looks good! :shipit:

PR #698, opened 2018-04-20
Add some NLE-like key bindings (similar to YouTube's)

rcombs:


mseeley:
Hey @rcombs I'm going to close this one out. We've got plenty of other ways to play/pause/seek.

The UI will seek when pressing left or right arrows and toggle play_pause using the space bar. Please reopen if you feel strongly that we should integrate.

Issue #695, opened 2018-04-18
player: Full-Screen mode for Plex Media Player not working correctly on macOS


paulallison99:
2 users are complaining about this issue on the this thread ( https://forums.plex.tv/discussion/307347/full-screen-mode-for-plex-media-player-not-working-correctly-on-macos-high-sierra ) and say it's still not fixed.

### Test environment
**PMS Version:** 1.11.2.4772 on macOS High Sierra
**PMP Version:** 2.4.1.787
**Platform (Windows/OSX/Embedded):** OSX

### Steps to reproduce
1. Open Plex Media Player in Desktop Mode. 
![image](https://user-images.githubusercontent.com/33132075/38945085-68cc81c6-432d-11e8-9072-13a3d015f158.png)

2. Click the green fullscreen button to make it go into full-screen mode. 
![image](https://user-images.githubusercontent.com/33132075/38945110-7af6c78a-432d-11e8-9d1e-2b1d7fbd7e8a.png)

![image](https://user-images.githubusercontent.com/33132075/38945115-8207b4da-432d-11e8-9947-ffc44ba9f5b5.png)

3. Quite out with Cmd+Q.

4. Open Plex Media Player and it will try to open but then fail and do a weird animation. (There is where the attached Gif comes in.) 

### Current behavior
The application does a weird minimize animation and then doesn't allow full-screen mode. There is an animation of this at https://forums.plex.tv/discussion/307347/full-screen-mode-for-plex-media-player-not-working-correctly-on-macos-high-sierra

### Expected behavior
Should go right into full-screen mode just fine and have no problems.

LongChair:
Probably related to https://github.com/plexinc/plex-media-player/issues/688

torarnv:
I've triaged this issue on the Qt side and I'm working on a fix upstream

LongChair:
Thanks @torarnv , nice to see feedback from Qt people :)

LongChair:
@paulallison99 : can you check if that build fixes the issue ?

https://www.dropbox.com/s/6154vjdoc4v6ori/PlexMediaPlayer-2.10.0.854-c1633f57-macosx-x86_64.zip?dl=0

kicjow:
@LongChair I just tested it on mine and it works perfectly now. 

BTW I am the one who made the thread that @paulallison99 posted.

LongChair:
@kicjow : I am glad to hear this is fixed.
Thanks for the feedback :)

Will close this when fix is merged

[...]

Issue #688, opened 2018-04-15
macOS - Window not focussed after opening PMP in full-screen mode


generator85:
### Test environment
**PMS Version:** 1.12.3.4973
**PMP Version:** 2.8.0.813-6e77dc0c
**Platform (Windows/OSX/Embedded):** MacOS 10.13.4

### Steps to reproduce
1. Open the app in full-screen mode
2. The app doesn't have focus. You'll have to click the window to be able to control it with remote or keyboard.

LongChair:
I have just tested this, and i don't seem to have the issue on OSX.
Starting PMP in fullscreen & using keyboard seems to work without any issue.

Could you check what has the focus when you start it ?

baz1860:
So, I have this issue too, but it only manifests when another app other than Finder has focus when I invoke PMP, and only when I invoke it from Spotlight.

generator85:
I've tested it with a script that shows what app has focus. When starting PMP it gains focus for a sec, after which focus is quickly returned to whatever app had focus before (e.g. Finder, Terminal). 

How I start PMP (like baz1860 describes) doesn't seem to matter in my case.

LongChair:
I just grabbed the 2.8.0 from plex.tv and started that one and still can't repro such issue.
My iMac is not quite up to date (10.12.6), but still .

Wether i start PMP from finder or spotlight doesn't seem to have any incidence.

Will update just in case some recent update broke it ...

LongChair:
Ok i have been able to reproduce this upgrading to High sierra.
It seems something with high sierra is breaking the Fullscreen mode on startup for a Qt application.
I was able to reproduce this in a Qt sample. Opening a Qt ticket here : https://bugreports.qt.io/browse/QTBUG-67840

torarnv:
This fix for this will be in 5.9.x, are you still on 5.7?

LongChair:
no we're using 5.9 :)

torarnv:
Ah, good! :) The deps in `README.md` could use an update then :)

LongChair:
indeed :)

torarnv:
You can cherry-pick https://codereview.qt-project.org/#/c/227619/ to very that the fix works for Plex

LongChair:
@torarnv thanks, i will try this. 
Meanwhile you're here though i'd have a question.
We have been having historically a lot of issues in out app with QQuickWindow and Fullscreen.

[...]

PR #699, opened 2018-04-20
macOS: let Cocoa handle window restoration

rcombs:
This fixes a number of quirks around startup where we'd fail to restore our previous geometry, particularly if entering fullscreen. Cocoa handles this much more reliably internally.

### DIFF
--- a/src/display/osx/DisplayManagerOSX.cpp
+++ b/src/display/osx/DisplayManagerOSX.cpp
@@ -51,7 +51,7 @@ bool DisplayManagerOSX::initialize()
     for (int modeid = 0; modeid < numModes; modeid++)
     {
       totalModes++;
-      
+
       // add the videomode to the display
       DMVideoModePtr mode = DMVideoModePtr(new DMVideoMode);
       mode->m_id = modeid;
@@ -94,7 +94,7 @@ bool DisplayManagerOSX::setDisplayMode(int display, int mode)
 {
   if (!isValidDisplayMode(display, mode) || !m_osxDisplayModes[display])
     return false;
-  
+
   CGDisplayModeRef displayMode =
   (CGDisplayModeRef)CFArrayGetValueAtIndex(m_osxDisplayModes[display], mode);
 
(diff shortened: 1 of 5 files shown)


LongChair:
i'll try this locally but could fix those
https://github.com/plexinc/plex-media-player/issues/688
https://github.com/plexinc/plex-media-player/issues/695

briankendall:
@rcombs Just built and tried out this PR, and PMP's window position and full screen state are not restored if I quit and relaunch the app. Running macOS 10.13.3.

LongChair:
That is also what i found out .. :/

LongChair:
PRs are welcome thanks !

I also dogged that today and found out some threads from other apps mentionning a workaround that would be to set window back to normal (rather than fullscreen) upon app closing.
I still need to check this.

LongChair:
Also another comment as I have seen this in several reports and I am not sure where this comes from.
We are using Qt fullscreen API. Qt is multi platform and I would expect it to use the respective and proper api for each platform.
I didn’t check it within Qt code but I assume that’s the way it should be ...

briankendall:
@LongChair I just dug into the full screen bugs, and found that Qt *does not* handle full screen properly in macOS, so fundamentally I think #688 and #695 are due to a Qt bug. The fix is to just use macOS's native API for toggling full screen.

LongChair:
The best thing to do would probably be to check what needs to be changed for Qt to support native OSX fullscreen API and PR that to Qt :)

briankendall:
I agree, but that strikes me as opening a can of worms! Maybe at some point when I have a good chunk of free time and am feeling a little masochistic I'll do that. And in the meantime I made #706.

LongChair:
Hehe it’s easier than what it looks like.
And I could help you there in the process as I landed a few up there already :)

I will have a look at your PR, thanks for submitting this :)
We still need to ponder if we want to add that workaround in PMP or get this fixed right upstream ;)

LongChair:
Anyways such PRs are very welcome, thanks once again !
I have seen a few minor things that I will comment out tomorrow :)

LongChair:
One more good reason to upstream fixes :)

mseeley:
January came and went. I'm closing this out due to overall inactivity.

PR #700, opened 2018-04-20
PlayerComponent: force seek support in all media

rcombs:
This allows Web to do some smarter things around MKV seeking.


LongChair:
I'm not sure what this is changing, could you tell me what kind of behavior this is affecting ? :)

rcombs:
Currently this shouldn't affect any actual behavior. It just allows Web to seek within the cache in a transcode, which it doesn't currently do, but should.

rcombs:
I'm comfortable merging this as-is.

PR #706, opened 2018-04-24
Use native full screen in macOS, fixes #688 and #695

### LINKED ISSUE #688: macOS - Window not focussed after opening PMP in full-screen mode

briankendall:
Qt's method for making a window full screen (i.e. `QWindow::setVisible(QWindow::FullScreen)`) is broken in macOS, in that it does not use macOS's native full screen functionality. This PR works around this by using Cocoa to make a window full screen natively when compiling for macOS. Fixes #688 and #695.

### DIFF
--- a/src/ui/KonvergoWindow.cpp
+++ b/src/ui/KonvergoWindow.cpp
@@ -106,7 +106,7 @@ KonvergoWindow::KonvergoWindow(QWindow* parent) :
 #endif
 
 #ifdef KONVERGO_OPENELEC
-  setVisibility(QWindow::FullScreen);
+  setWindowFullScreen(true);
 #else
   updateWindowState(false);
 #endif
@@ -427,25 +427,31 @@ void KonvergoWindow::updateMainSectionSettings(const QVariantMap& values)
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
-void KonvergoWindow::updateForcedScreen()
+bool KonvergoWindow::updateForcedScreen()
 {
   QString screenName = SettingsComponent::Get().value(SETTINGS_SECTION_MAIN, "forceFSScreen").toString();
 
   if (screenName.isEmpty())
-    return;
+    return false;
 
   for (QScreen* scr : QGuiApplication::screens())
   {
     if (scr->name() == screenName)
     {
       QLOG_DEBUG() << "Forcing screen to" << scr->name();
+#ifdef Q_OS_MAC
+      OSXUtils::SetWindowFullScreenOnSpecificScreen(this, scr->geometry());
+#else
       setScreen(scr);
       setGeometry(scr->geometry());
-      setVisibility(QWindow::FullScreen);
+      setWindowFullScreen(true);
+#endif
       InputComponent::Get().cancelAutoRepeat();
-      return;
+      return true;
     }
   }
+  
+  return false;
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
@@ -458,19 +464,36 @@ void KonvergoWindow::updateWindowState(bool saveGeo)
     if (!isFullScreen() && saveGeo)
       saveGeometry();
 
-    setVisibility(QWindow::FullScreen);
-
     // When fullscreening explicitly, we might have to move the window to a
-    // different screen, as Qt will fullscreen to the current screen.
+    // different screen, as both Qt and macOS will fullscreen to the current
+    // screen.
+#ifdef Q_OS_MAC
+    if (!updateForcedScreen()) {
+      setWindowFullScreen(true);
+    }
+#else
     QTimer::singleShot(200, [=]
     {
       updateForcedScreen();
     });
+#endif
   }
   else
   {
-    setVisibility(QWindow::Windowed);
+    setWindowFullScreen(false);
+    
+#ifdef Q_OS_MAC
+    // We have to wait for macOS's full screen animation to finish before we
+    // move the window, otherwise it may move to the other screen before the
+    // window server recognizes it's actually on that screen, causing it to
+    // be completely invisible and with essentially no way to get it back!
+    QTimer::singleShot(1000, [=]
+    {
+      loadGeometry();
+    });
+#else
     loadGeometry();
+#endif
 
     Qt::WindowFlags forceOnTopFlags = Qt::WindowStaysOnTopHint;
 #ifdef Q_WS_X11
@@ -520,7 +543,7 @@ void KonvergoWindow::onVisibilityChanged(QWindow::Visibility visibility)
     {
       QLOG_WARN() << "winging it!";
       setScreen(realScreen);
-      setVisibility(QWindow::FullScreen);
+      setWindowFullScreen(true);
       return;
     }
   }
@@ -530,7 +553,7 @@ void KonvergoWindow::onVisibilityChanged(QWindow::Visibility visibility)
   {
     QLOG_WARN() << "Forcing re-entering fullscreen because of forceAlwaysFS setting!";
     updateForcedScreen(); // if a specific screen is forced, try to move the window there
-    setVisibility(QWindow::FullScreen);
+    setWindowFullScreen(true);
     return;
   }
 
@@ -772,4 +795,23 @@ void KonvergoWindow::updateCurrentScreen()
   QString currentName = current ? current->name() : "";
   if (currentName != m_currentScreenName)
     updateScreens();
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void KonvergoWindow::setWindowFullScreen(bool fullScreen)
+{
+#ifdef Q_OS_MAC
+    if (fullScreen) {
+        setVisibility(QWindow::Windowed);
+        OSXUtils::SetWindowFullScreen(this, true);
+    } else {
+        OSXUtils::SetWindowFullScreen(this, false);
+        setVisibility(QWindow::Windowed);
+    }
+#else
+    if (fullScreen)
+        setVisibility(QWindow::FullScreen);
+    else
+        setVisibility(QWindow::Windowed);
+#endif
 }
--- a/src/ui/KonvergoWindow.h
+++ b/src/ui/KonvergoWindow.h
@@ -122,8 +122,9 @@ private:
   bool fitsInScreens(const QRect& rc);
   QScreen* loadLastScreen();
   void updateScreens();
-  void updateForcedScreen();
+  bool updateForcedScreen();
   QScreen* findCurrentScreen();
+  void setWindowFullScreen(bool fullScreen);
 
   bool m_debugLayer;
   QTimer* m_infoTimer;
--- a/src/utils/osx/OSXUtils.mm
+++ b/src/utils/osx/OSXUtils.mm
@@ -1,5 +1,7 @@
 #include "OSXUtils.h"
 #include "QsLog.h"
+#include "ui/KonvergoWindow.h"
+#include <QTimer>
 #import <Cocoa/Cocoa.h>
 
 /////////////////////////////////////////////////////////////////////////////////////////
@@ -80,4 +82,84 @@ void OSXUtils::SetCursorVisible(bool visible)
     [NSCursor unhide];
   else
     [NSCursor hide];
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static NSWindow * qtWindowToCocoaWindow(QWindow *qtWindow)
+{
+  if (!qtWindow)
+    return nil;
+
+  NSView* cocoaView = (NSView*)qtWindow->winId();
+
+  if (!cocoaView)
+    return nil;
+
+  return cocoaView.window;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static bool isWindowFullScreen(NSWindow *cocoaWindow)
+{
+  return cocoaWindow.styleMask & NSFullScreenWindowMask;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+bool OSXUtils::isWindowFullScreen(QWindow *qtWindow)
+{
+  NSWindow* cocoaWindow = qtWindowToCocoaWindow(qtWindow);
+
+  if (!cocoaWindow)
+    return false;
+
+  return isWindowFullScreen(cocoaWindow);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static void SetWindowFullScreen(NSWindow *cocoaWindow, bool fullScreen)
+{
+  if (!cocoaWindow || fullScreen == isWindowFullScreen(cocoaWindow))
+    return;
+
+  [cocoaWindow toggleFullScreen:nil];
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreen(QWindow *qtWindow, bool fullScreen)
+{
+  SetWindowFullScreen(qtWindowToCocoaWindow(qtWindow), fullScreen);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static void MoveWindowWithinScreen(NSWindow *cocoaWindow, QRect screenRect)
+{
+  if (!cocoaWindow)
+    return;
+
+  int padding = 50;
+  [cocoaWindow setFrame:NSMakeRect(screenRect.left() + padding, screenRect.top() + padding,
+                                   screenRect.width() - padding*2, screenRect.height() - padding*2) display:YES animate:NO];
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreenOnSpecificScreen(QWindow *qtWindow, QRect screenRect)
+{
+  NSWindow* cocoaWindow = qtWindowToCocoaWindow(qtWindow);
+
+  if (!cocoaWindow)
+    return;
+
+  if (isWindowFullScreen(cocoaWindow)) {
+    SetWindowFullScreen(cocoaWindow, false);
+
+    // Need to wait for the full screen animation to finish
+    QTimer::singleShot(1000, [=] ()
+    {
+      MoveWindowWithinScreen(cocoaWindow, screenRect);
+      SetWindowFullScreen(cocoaWindow, true);
+    });
+  } else {
+    MoveWindowWithinScreen(cocoaWindow, screenRect);
+    SetWindowFullScreen(cocoaWindow, true);
+  }
 }
--- a/src/utils/osx/OSXUtils.h
+++ b/src/utils/osx/OSXUtils.h
@@ -2,6 +2,8 @@
 #define OSXUTILS_H
 
 #include <QString>
+#include <QWindow>
+#include <QRect>
 #include <ApplicationServices/ApplicationServices.h>
 
 namespace OSXUtils
@@ -13,6 +15,9 @@ namespace OSXUtils
   unsigned long GetPresentationOptions();
   unsigned long GetPresentationOptionsForFullscreen(bool hideMenuAndDock);
   void SetCursorVisible(bool visible);
+  bool isWindowFullScreen(QWindow *window);
+  void SetWindowFullScreen(QWindow *window, bool fullScreen);
+  void SetWindowFullScreenOnSpecificScreen(QWindow *qtWindow, QRect screenRect);
 };
 
 #endif /* OSXUTILS_H */


briankendall:
Also I should add:

In macOS it's possible to switch the PMP window from windowed to full screen using the native green zoom button. This button is available when PMP is in full screen mode, as is standard in macOS.

The way I've implemented this PR means that clicking the zoom button will not change the "full screen" option in PMP's settings, so if you change full screen mode only using the native zoom button and then quit the app, upon next launch it will continue to respect the full screen setting rather than remember which state it was in when it quit.

Me and a friend agree that this is how PMP should behave: the full screen setting should remain unchanged regardless of how the user manipulates the window using the native zoom button. However, if anyone on the dev team thinks this behavior is incorrect (and I think a solid argument could be made that it is), I can amend the PR so that the "full screen" setting will change when the app's full screen state is changed natively.

src/ui/KonvergoWindow.cpp
```diff
@@ -106,7 +106,7 @@ KonvergoWindow::KonvergoWindow(QWindow* parent) :
 #endif
 
 #ifdef KONVERGO_OPENELEC
-  setVisibility(QWindow::FullScreen);
+  setWindowFullScreen(true);
 #else
```
LongChair: For the window method, `setFullScreen` name should be preferred.

src/ui/KonvergoWindow.cpp
```diff
@@ -458,18 +458,19 @@ void KonvergoWindow::updateWindowState(bool saveGeo)
     if (!isFullScreen() && saveGeo)
       saveGeometry();
 
-    setVisibility(QWindow::FullScreen);
-
+    setWindowFullScreen(true);
+#ifndef Q_OS_MAC
     // When fullscreening explicitly, we might have to move the window to a
```
LongChair: I'm not sure this is valid to ifdef that here. There are settings for the physicaly screen that the windows was last fullscreened on which is what `updateForcedScreen`will try to restore.
Native OSX fullscreen will indeed send it to a new desktop but that will remain on the same screen.
We should make sure that this behaves the same on OSX, so probably better to have the same scheme, if possible.
if you have 2 screens on OSX and decide PMP should be on second screen, the window has to restore fullscreen on second screen.

src/ui/KonvergoWindow.cpp
```diff
@@ -441,7 +441,7 @@ void KonvergoWindow::updateForcedScreen()
       QLOG_DEBUG() << "Forcing screen to" << scr->name();
       setScreen(scr);
       setGeometry(scr->geometry());
-      setVisibility(QWindow::FullScreen);
+      setWindowFullScreen(true);
```
LongChair: naming

src/ui/KonvergoWindow.cpp
```diff
@@ -458,18 +458,19 @@ void KonvergoWindow::updateWindowState(bool saveGeo)
...
     QTimer::singleShot(200, [=]
     {
       updateForcedScreen();
     });
+#endif
   }
   else
   {
-    setVisibility(QWindow::Windowed);
+    setWindowFullScreen(false);
```
LongChair: naming everywhere else :)

src/ui/KonvergoWindow.cpp
```diff
@@ -773,3 +774,22 @@ void KonvergoWindow::updateCurrentScreen()
   if (currentName != m_currentScreenName)
     updateScreens();
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void KonvergoWindow::setWindowFullScreen(bool fullScreen)
```
LongChair: naming here, but makes sense to keep the name you gave for the `OSXUtils` one

src/ui/KonvergoWindow.cpp
```diff
@@ -773,3 +774,22 @@ void KonvergoWindow::updateCurrentScreen()
...
+{
+#ifdef Q_OS_MAC
+    if (fullScreen) {
```
LongChair: You can probably make this block simpler with 
```
        setVisibility(QWindow::Windowed);
        OSXUtils::SetWindowFullScreen(this, fullscreen);
```

src/ui/KonvergoWindow.h
```diff
@@ -124,6 +124,7 @@ private slots:
   void updateScreens();
   void updateForcedScreen();
   QScreen* findCurrentScreen();
+  void setWindowFullScreen(bool fullScreen);
```
LongChair: naming

src/utils/osx/OSXUtils.mm
```diff
@@ -1,5 +1,6 @@
 #include "OSXUtils.h"
 #include "QsLog.h"
+#include "ui/KonvergoWindow.h"
```
LongChair: We can more likely just use the QWindow class include here rather than using our specific window, would be more generic. ie `#include <QWindow>`

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +82,22 @@
   else
     [NSCursor hide];
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////
+bool OSXUtils::isWindowFullScreen(KonvergoWindow *qtWindow)
```
LongChair: Change type to `QWindow *` here, also nust name it `window` , type makes its explicit that it's a Qt one :)

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +82,22 @@
...
+{
+  NSView* cocoaView = (NSView*)qtWindow->winId();
```
LongChair: nullcheck for window

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +82,22 @@
...
+  return cocoaWindow.styleMask & NSFullScreenWindowMask;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreen(KonvergoWindow *qtWindow, bool fullScreen)
+{
+  if (fullScreen == isWindowFullScreen(qtWindow))
+    return;
+
+  NSView* cocoaView = (NSView*)qtWindow->winId();
```
LongChair: nullcheck for window
you also most likely want to make sure  you get a cocoaView valid pointer here ?

LongChair:
@briankendall : made a few comments, also please make sure you read https://github.com/plexinc/plex-media-player/blob/master/CONTRIBUTING.md and adjust this according to it. ie especially the CLA part & commit message :)

Thanks.

src/ui/KonvergoWindow.cpp
briankendall: `setFullScreen` is already in use by another method, though, and it serves a different purpose of updating the application setting. What name should I use in light of that?

src/ui/KonvergoWindow.cpp
briankendall: I was thinking that it would be best to use the default macOS behavior, but I can make that change.

src/utils/osx/OSXUtils.mm
briankendall: Good call

briankendall:
@LongChair Would you prefer I amend my current commit or create another one with these changes?

LongChair:
amending is fine :)

src/ui/KonvergoWindow.cpp
LongChair: Good point, i guess that name is fine then, at least i have no better idea.

src/ui/KonvergoWindow.cpp
LongChair: yeah i think if we have options, they just need to be effective.

briankendall: Okay, I've been bashing my head against this, and I'm starting to get the feeling that it's not possible to implement this feature in macOS, at least not in a way that will work consistently. Fundamentally the issue is that full screen mode is a different beast in macOS than in any other OS. It has it's own standard behaviors and workflows, and one of the biggest fundamental differences is that apps are not supposed to have any control over which monitor they will occupy in full screen mode, nor are they supposed to have the ability to move between monitors. So we are very much going against the grain in implementing a feature like this.

So trying to control which monitor the window will use when it becomes full screen is unfortunately not a feature available in the macOS API. By and large a window will become full screen on whichever monitor it's on, but it's only guaranteed to work if the user drags the window to that monitor themselves.

If I use any of a variety of API calls to move the window to the correct monitor, *sometimes* macOS will fail to register that the window has changed screens. This has two bad effects: telling it to go full screen will of course send it to the wrong monitor, and worse yet, the monitor will be invisible with no easy way to bring it back. I cannot for the life of me figure out why this is happening, why it works sometimes but not other times, or whether there's a solution. It looks like a macOS window server bug, or more likely it might be something bad that Qt is doing in their implementation of `NSWindow`.

However, the one place this *does* seem to work consistently is when PMP starts up. So maybe it's worth keeping this in just so that on launch it will start on the correct display.

But there are a number cases where we simply cannot control which monitor it will use for full screen mode:

- The user can exit full screen mode using the title bar and drag the window to another monitor, as is normal in macOS
- The user can use Mission Control to move the window between monitors, as is normal in macOS
- If the user exits full screen mode via the app settings and then re-enables full screen mode later, we cannot ensure the window will appear on the correct monitor.
- Changing the "Force fullscreen display" setting generally does not successfully move the window to another monitor, though after rebooting PMP the window *will* appear on the correct monitor

I'm not sure how you think it's best to handle this. I'm thinking that in macOS PMP could respect the "force fullscreen display" setting at start up, but past that it will not try to force the window to appear on a particular display. Instead the user will have to move the window themselves via Mission Control or some other method (which is how things are normally done in macOS).

What do you think?

briankendall:
Okay, just finished a first pass at the requested changes. However, as mentioned in one of my comments, the "Force fullscreen display" does not work consistently, and probably cannot ever work consistently in macOS. I wrote about that in a lot more detail in the change request thread. I'm curious what you all think is the correct approach.

src/ui/KonvergoWindow.cpp
briankendall: Just to make this even more complicated, I just tried running the last build of PMP I made a few hours ago and suddenly "Force fullscreen display" is working correctly. Still not sure what causes it to get into a broken state, nor how often it happens!

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static NSWindow * qtWindowToCocoaWindow(QWindow *qtWindow)
+{
+  if (!qtWindow)
+    return nil;
+
+  NSView* cocoaView = (NSView*)qtWindow->winId();
+
+  if (!cocoaView)
```
sixones: Use `cocoaView == nil`.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
   else
     [NSCursor hide];
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static NSWindow * qtWindowToCocoaWindow(QWindow *qtWindow)
+{
+  if (!qtWindow)
```
sixones: Use `qtWindow == nil`.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+{
+  return cocoaWindow.styleMask & NSFullScreenWindowMask;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+bool OSXUtils::isWindowFullScreen(QWindow *qtWindow)
+{
+  NSWindow* cocoaWindow = qtWindowToCocoaWindow(qtWindow);
+
+  if (!cocoaWindow)
```
sixones: Use `cocoaWindow == nil`.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+    return false;
+
+  return isWindowFullScreen(cocoaWindow);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static void SetWindowFullScreen(NSWindow *cocoaWindow, bool fullScreen)
+{
+  if (!cocoaWindow || fullScreen == isWindowFullScreen(cocoaWindow))
```
sixones: Use `cocoaWindow == nil` again.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreen(QWindow *qtWindow, bool fullScreen)
+{
+  SetWindowFullScreen(qtWindowToCocoaWindow(qtWindow), fullScreen);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+static void MoveWindowWithinScreen(NSWindow *cocoaWindow, QRect screenRect)
+{
+  if (!cocoaWindow)
```
sixones: Same again use `== nil`.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+  [cocoaWindow setFrame:NSMakeRect(screenRect.left() + padding, screenRect.top() + padding,
+                                   screenRect.width() - padding*2, screenRect.height() - padding*2) display:YES animate:NO];
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreenOnSpecificScreen(QWindow *qtWindow, QRect screenRect)
+{
+  NSWindow* cocoaWindow = qtWindowToCocoaWindow(qtWindow);
+
+  if (!cocoaWindow)
```
sixones: Same here, use `cocoaWindow == nil`.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+  if (isWindowFullScreen(cocoaWindow)) {
+    SetWindowFullScreen(cocoaWindow, false);
+
+    // Need to wait for the full screen animation to finish
+    QTimer::singleShot(1000, [=] ()
+    {
+      MoveWindowWithinScreen(cocoaWindow, screenRect);
+      SetWindowFullScreen(cocoaWindow, true);
+    });
+  } else {
```
sixones: Brackets should be on new lines.

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+void OSXUtils::SetWindowFullScreenOnSpecificScreen(QWindow *qtWindow, QRect screenRect)
+{
+  NSWindow* cocoaWindow = qtWindowToCocoaWindow(qtWindow);
+
+  if (!cocoaWindow)
+    return;
+
+  if (isWindowFullScreen(cocoaWindow)) {
```
sixones: Bracket should be on new line.

src/ui/KonvergoWindow.cpp
```diff
@@ -773,3 +796,22 @@ void KonvergoWindow::updateCurrentScreen()
...
+        setVisibility(QWindow::Windowed);
+        OSXUtils::SetWindowFullScreen(this, true);
+    } else {
```
sixones: New lines for backets.

src/ui/KonvergoWindow.cpp
sixones: New lines for brackets again.

briankendall:
Okay, I can make these changes in the not too distant future. Any thoughts on the unreliability of programmatically switching a full screen window to a different monitor?

Here's the write up I did from an out of date comment thread:

> Okay, I've been bashing my head against this, and I'm starting to get the feeling that it's not possible to implement this feature in macOS, at least not in a way that will work consistently. Fundamentally the issue is that full screen mode is a different beast in macOS than in any other OS. It has it's own standard behaviors and workflows, and one of the biggest fundamental differences is that apps are not supposed to have any control over which monitor they will occupy in full screen mode, nor are they supposed to have the ability to move between monitors. So we are very much going against the grain in implementing a feature like this.
> 
> So trying to control which monitor the window will use when it becomes full screen is unfortunately not a feature available in the macOS API. By and large a window will become full screen on whichever monitor it's on, but it's only guaranteed to work if the user drags the window to that monitor themselves.
> 
> If I use any of a variety of API calls to move the window to the correct monitor, sometimes macOS will fail to register that the window has changed screens. This has two bad effects: telling it to go full screen will of course send it to the wrong monitor, and worse yet, the monitor will be invisible with no easy way to bring it back. I cannot for the life of me figure out why this is happening, why it works sometimes but not other times, or whether there's a solution. It looks like a macOS window server bug, or more likely it might be something bad that Qt is doing in their implementation of NSWindow.
> 
> However, the one place this does seem to work consistently is when PMP starts up. So maybe it's worth keeping this in just so that on launch it will start on the correct display.
> 
> But there are a number cases where we simply cannot control which monitor it will use for full screen mode:
> 
> * The user can exit full screen mode using the title bar and drag the window to another monitor, as is normal in macOS
> * The user can use Mission Control to move the window between monitors, as is normal in macOS
> * If the user exits full screen mode via the app settings and then re-enables full screen mode later, we cannot ensure the window will appear on the correct monitor.
> * Changing the "Force fullscreen display" setting generally does not successfully move the window to another monitor, though after rebooting PMP the window will appear on the correct monitor
> 
> I'm not sure how you think it's best to handle this. I'm thinking that in macOS PMP could respect the "force fullscreen display" setting at start up, but past that it will not try to force the window to appear on a particular display. Instead the user will have to move the window themselves via Mission Control or some other method (which is how things are normally done in macOS).
> 
> What do you think?

src/utils/osx/OSXUtils.mm
```diff
@@ -81,3 +83,83 @@
...
+    SetWindowFullScreen(cocoaWindow, false);
+
+    // Need to wait for the full screen animation to finish
+    QTimer::singleShot(1000, [=] ()
```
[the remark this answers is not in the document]
briankendall: Generally it should be reliable, as the animation usually takes less than 0.7 seconds to complete, and the extra ~0.3 seconds give us a little head room. However, if we wanted to be more explicit about it, it's possible to receive notifications for when the full screen transition has completed.

mseeley:
Agreed, thank you @briankendall.

To me, restoring fullscreen at startup is a must for PMP when used on an appliance-like HTPC. The restoration, to me (again), loses value during desktop usage.

I'd like us to be in a place where the PMP window has no uncanny behaviors. Keyboard window handling shortcuts and OS paradigms should be 100% supported. If our implementation is going against the grain let's quit that immediately.

I would like us to consider:

* Re-evaluating window handing after #688 resolves. Huge thanks to @torarnv for joining us here.
* Remove our code which attempts to force PMP to be aware of specific displays.
  * Is it really helpful outside of truly niche setups?
  * As a user is it easier to move the window to a different monitor than it is to recover from a bad monitor selection?
* Allow the fullscreen toggle to take effect at startup or during app usage; allow the OS to select the display.
* Investigate fullscreen restoration on Windows and macOS. Does the OS treat exiting an application in fullscreen mode as an invitation to restore fullscreen mode at next app start?
  * Personally I don't feel there's a case to restore to a fullscreen window _unless_ the user has configured that setting manually.

briankendall:
Would you guys like me to continue working on this PR for the immediate future? Based on the conversation in #688 it sounds like PMP is going to switch over to a later version of Qt that uses native full screen in macOS, which covers about half of what this PR currently does. Also there's the discussion happening here about how we actually want this feature to work. It might make more sense for me to wait for some of these to come to fruition and then redo parts of this PR.

> To me, restoring fullscreen at startup is a must for PMP when used on an appliance-like HTPC. The restoration, to me (again), loses value during desktop usage.

I agree. For certain use cases this is critical!

> I'd like us to be in a place where the PMP window has no uncanny behaviors. Keyboard window handling shortcuts and OS paradigms should be 100% supported. If our implementation is going against the grain let's quit that immediately.

Agree here too. I think there's a nice middle ground here where we can have PMP restore to full screen on a particular monitor at launch, but otherwise allow the app to work as other macOS apps do where the user can toggle full screen mode and/or manipulate the window however they like.

Also worth noting that macOS *does* have a facility for restoring windows to their previous position / full screen state, and @rcombs started #699 for that. However I'm not sure if we should rely on it exclusively. It can only be enabled or disabled for the entire system and it comes with a lot of other behavior (e.g. how macOS handles closing unsaved documents). I prefer to have it disabled, but would still like for PMP to restore to its previous state.

LongChair:
@briankendall : my thoughts are as follows :grinning: 
Let's try to see how the fix provided by Qt is making it behave and how much of our current issues it's solving. I made overnight a Qt 5.9.5 build including the patch. and I'm currently working on making a build out of this.
I agree with you that restoring fullscreen on startup is important for people using it on their TV where they don't have a mouse on hand to click the fullscreen button. that is bound to a setting that is handled by PMP. the state of this window is saved everytime user changes the window state so that when he exits / return, the previous window state is restored.
I don't think this is such a wierd behavior and i don't see how that would be incompatible with keeping the ability of toggling the fullscreen OSX button on top of the window, but i could be wrong.
So i'll share that build on the GHI issue and ping you all there when i have completed it.
We'll tackle this one up to death, because i'm tired of those issues which have been standing there for too long :)
@briankendall : so hold on on this PR, lets see what Qt fixes or not, and then we can adjust anything that's required :)

mseeley:
Hi folks. I'm going to close this out. After we get an updated Qt integrated we can sort out any remaining window weirdness. Objectively the PMP window should be very careful when handling concerns in place of the system's window manager.

Issue #685, opened 2018-04-10
macOS - PMP loses all window widgets when shrinking from ⌘F


ajslater:
- If you maximize PMP with ⌘F it maximizes correctly, but loses all its window widgets. which doesn't matter because you can't do anything with a maximized window.
- But If you then un-maximize PMP with ⌘F, PMP shrinks to a window size but lacks all window widgets. It cannot be dragged or controlled by the macOS window manager. There is no window border or window header.

However, if you maximize PMP with the green maximize button it retains its window widgets and borders and can be un-maximized correctly.

I think the bug happens with ⌘F maximizing, it does something different than maximizing with the green button and destroys the window widgets.

kinglouie:
can confirm this issue, PMPs fullscreen behaves very weird. It seems it doesn't use macOS native Fullscreenapi

ajslater:
Possibly related to https://github.com/plexinc/plex-media-player/issues/664

ajslater:
Sorry, my bad. It used to be Cmd F. It was changed to Cmd-Shift-F (for some reason I don't fully agree with), but the behavior is the same as I described.

LongChair:
@ajslater @kinglouie : can you please check if that build fixed the issue ? 

https://www.dropbox.com/s/6154vjdoc4v6ori/PlexMediaPlayer-2.10.0.854-c1633f57-macosx-x86_64.zip?dl=0

ajslater:
Yes it keeps all window border widgets when maximizing and un-maximizing
with Cmd-Shift-F. Huzzah! :D Thanks.

Bug with this build:
The Movie index display contains some duplicate entries with this build.
Using the current Plex Plus Channel build these ghost duplicates do not
appear.
Underneath most movies the year of screening appears. Underneath the
duplicate movies in the experimental build the subscript "1 movie" or
occasionally "2 movies" appears. Clicking on them just plays the regular
movie. The movies with duplicates do not seem special in any way, they do
not have extras or alternate versions.

kinglouie:
This build fixes the issue for me as well. 👍

LongChair:
PR addressing this issue was merged. Closing :)

Issue #677, opened 2018-03-26
Major issues with intel graphics chips and display mode switching


Zoratt:
A link to information from quite a few users. https://forums.plex.tv/discussion/294977/problem-with-display-mode-switching

When you watch a video that the gpu tries to change to 24fps, the screen blacks out and everything freezes.

Since the Creators version of Windows 10 was forced on us, everything was working fine.

The Open GL can get us to limp on for a while, but it has continued to be an issue since November of 2017 and new drivers, and updates to PMP haven't addressed the issue.

Flippage:
Same issue here. We have to alt+tab out and back in when we play or exit any video. And if I roll back my windows version, it automatically installs the updates again.

tommeh1337:
I have the exact same problem. PMP + W10 Fall creators update + display mode switching is a no-go.

Morfeu78:
I have the same problem with display switching.  Hardware is a Nuc 54250 + Marantz 5011 + LG 65e6. Software is PMP + W10 Fall Creators Update.

Morfeu78:
P.s.: open glass does not resolve the issue for me.

Morfeu78:
*GL

LongChair:
Yeah the W10 Fall creators update has been breaking ANGLE support it seems.
We discussed that internally to restore OpenGL as default rendering method in next updates.

Until then you can change this to OpenGL, it should fix the issue.

Flippage:
Changed to OpenGL. I also updated my Intel HD graphics drivers to the latest version. openGL seems to have done the trick, and I'm not getting the issues I used to get when running OpenGL. 

Happy camper here. 

Fyi - I did some googling, and I think this is a Windows 10 issue. Probably needs to be bugfixed by Microsoft.

adaigre:
Same problem here since 1709 ...
But I have ATI HD 6XXX (don't remember).
Really annoying.

MattyBalaam:
I have tried updating my drivers and switching on Native OpenGL but this does not fix the issue.

LongChair:
> but the screen resolution does not match the video.

I'm not sure to understand what you mean there, can you be a bit more specific ?
If you mean that AutoFrameRate will not switch to 720p video mode when you're running 1080p video mode, but just switches refresh rate, that is by design. a,d there is not much that can be done about this at the moment.

Thanks for you feedback.

adaigre:
@LongChair Sorry !
[...]

Issue #626, opened 2017-10-15
Interface does not fade away after mouse out in top/bottom of screen


tehsolace:
While not in full-screen mode, play a video.  If you mouse-out of the top or bottom edges of the player's window, the interface will not fade away after few seconds (bug).  If you mouse-out of the left or right edges of the player's window, the interface will fade away (expected) as normal.

Tested in Windows 10 with current version of Plex Media Player 3.21.2.

LongChair:
@tehsolace : could you confirm if this is still the case in current builds please ?

DarkNemie:
As of Build (64-bit) 2.10.0.849-e02dbeca May 1, 2018 Running on windows 10 Pro x64 with Radeon 270x and Intel i7 - 4770k cpu The issue still exists. one thing to note is mousing out through the bottom UI to slowly or quickly seems to cause it more. Also this issue is worsened when PmP is used in a Multi monitor setup such as triple monitors with the PMP in the monitor that is above the other 2. Also this Issues exists when PMP is both in windowed mode and in full screen mode, I have not tested this behaviour in TV mode though.

DarkNemie:
I have tested Build 12.10.0.854 that you posted on drop box. I can confirm in a Multi-Monitor setup where the previous versions failed to fade away, This one works great, the UI as well seems considerably more responsive as well. I have tested in windowed mode, fullscreen, windowed mode, Fullscreen mode. All are working as intended. The only thing I noticed is some times when moving the mouse into the plex monitor or window the UI will appear/disappear rapidly 3 or 4 times before staying, This does not affect anything as for usage just Thought it should be noted. Thank you. 

Also after a full 12 hours of usage I have not seen any other issues from the build.

LongChair:
Great and thanks for the feedback.
Fix was merged, will close it then.

Issue #599, opened 2017-08-21
NVidia Fast sync causes crash to desktop


Ca-pp:
In NVidia control Panel -> Manage 3D Settings -> Global Settings

Set Vertical Sync to 'Fast' and Apply

Open Plex Media Player and it will crash back to desktop after a moment.

Version 1.3.5
Windows 10
Nvidia Driver 385.28

[PlexMediaPlayer.Log.txt](https://github.com/plexinc/plex-media-player/files/1238224/PlexMediaPlayer.Log.txt)

spoonypirate:
I'll second this as still being an issue. Changing to (for example) Adaptive VSync allows fullscreen to work with Plex Media Player.

Windows 10
PMP Version 2.1.1
Nvidia Driver 384.76

rnyberg:
Just found this ticket and deactivated fast vsync, now my PMP no longer crashesh instantly, so I guess.. I hird this. ;)

LongChair:
Would you guys mind checking if https://github.com/plexinc/plex-media-player/issues/677#issuecomment-384927456 is making any difference for this Nvidia specific issue ?

Thansk to provide me with some feedback :)

LongChair:
@AndyHeap @redfellow @spoonypirate : could anyone of you please check with replacing the dlls i mentionned and let me know ?

redfellow:
I'll test the DLLs tonight if I have time for it, thanks for your effort.

spoonypirate:
@LongChair I checked out the mentioned build and I can go into fullscreen no problems with Fast Sync enabled.

LongChair:
@spoonypirate : awesome news :) thanks for checking !

LongChair:
@spoonypirate @redfellow : Here s the latest build which should incldue the fix.

https://www.dropbox.com/s/w7neafa3ymv86mk/PlexMediaPlayer-2.10.0.854-c1633f57-windows-x64.exe?dl=0

Please be patient until this land.

LongChair:
PR addressing this issue was merged. Closing :)

PR #713, opened 2018-04-28
Update to Qt 5.9.5 

LongChair:
Ok this is a nice PR that is not intended to be merged yet, but still openning it for discussion.

We have been having long standing displaying issues in PMP :

- With **OSX** and fullscreen issues
- With **Windows** due to Win10 creator update which was forced on us and broke angle API usage.

That PR Bumps Qt to 5.9.5 because we need a few recent fixes that have been brought there for OSX fullscreen. 
We also need a new Qt dependencies package because we will be using now the Qt packaged Angle dependencies rather than building our own seperately which makes them harder to maintain an inconsistent with Qt changes.

Regarding OSX, this PR also removes the OSX presentation options that completely remove the topbar & taskbar. this is not reliable when switching screens, and allows to restore OSX native behavior.

Issues impacted with this PR (ticked = confirmed, unticked = to be confirmed):
 - OSX :
    - [X] #695
    - [X] #688
    - [X] #685
    - [ ] #626  
    - [ ] #508
    - [ ] #503
    - [ ] #294
    - [ ] #210        

 - Windows :
    - [X] #677
    - [X] #599
    - [X] #715

 - Others
    - [ ] #527

### DIFF
--- a/CMakeModules/DependencyConfiguration.cmake
+++ b/CMakeModules/DependencyConfiguration.cmake
@@ -37,7 +37,7 @@ if(WIN32)
     ARCHSTR windows-x86_x64
   )
 
-  option(ENABLE_ANGLE_DEP "Enable downloading and deployment of separate ANGLE deps" ON)
+  option(ENABLE_ANGLE_DEP "Enable downloading and deployment of separate ANGLE deps" OFF)
   if(ENABLE_ANGLE_DEP)
     message(STATUS "Downloading separate ANGLE deps...")
     foreach(angle_file libEGL.dll libGLESv2.dll d3dcompiler_47.dll)
--- a/CMakeModules/QtConfiguration.cmake
+++ b/CMakeModules/QtConfiguration.cmake
@@ -10,7 +10,7 @@ if(NOT IS_DIRECTORY ${QTROOT})
 		"plexmediaplayer-qt"
 		DIRECTORY dir
 		DEPHASH_VAR QT_DEPS_HASH
-    BUILD_NUMBER 165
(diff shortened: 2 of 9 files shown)


torarnv:
I should note that Qt 5.9 is an LTS release, that will continue to see a few critical bug fixes, but I would recommend following the 5.9 upgrade with a bump to the soon to be released Qt 5.11 LTS. 

https://wiki.qt.io/Qt_5.9_Release
https://wiki.qt.io/Qt_5.11_Release

LongChair:
@torarnv : thanks for the advice. Qt 5.11 will require some further work for us, as it has requirements for VS2017 on windows. So far we need some CI changes to be able to build it properly.

LongChair:
Here is a test build for those willing to test : 

https://nightlies.plex.tv/plexmediaplayer/dist-qt-5.9.5/2.10.0.854-c1633f57/

LongChair:
@torarnv : yeah i will bump those requirements in the PR.

The qt 5.9.5 is maitained in a separate private repo. but i can add it to https://github.com/plexinc/plex-media-player/tree/master/qt-patches if that is what you mean :)

torarnv:
> The qt 5.9.5 is maitained in a separate private repo. but i can add it to https://github.com/plexinc/plex-media-player/tree/master/qt-patches if that is what you mean :)

Yupp, that's what I meant :) The current PMP repo will not build against vanilla Qt 5.9.5, so having the patch available to contributors and in the build instructions is important I think.

I would of course recommend just making the repo with the individual patches public :)

PR #727, opened 2018-05-08
Bump Mpv to 1.2

LongChair:
This PR bumps Mpv to 1.2 and aligns ffmpeg to PMS version

It also changes the embedding API to new Mpv render API.
The API we were using is deprecated now.

It has been tested by ninjas and does not show any issues so far


markus101 (approved):
Testing well, but I can speak for the code, I'll defer to @sixones and/or @rcombs on that.

src/player/PlayerComponent.cpp
```diff
@@ -630,12 +636,13 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
       }
       // Start "on_preloaded" hook.
       // Used initialize stream selections and to probe codecs.
-      if (!strcmp(msg->args[1], "2"))
+      if (!strcmp(hook->name, "on_preloaded"))
```
sixones: Not this PRs fault but should use `== false`.

src/player/PlayerComponent.cpp
```diff
@@ -630,12 +636,13 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
...
       {
         reselectStream(m_currentSubtitleStream, MediaType::Subtitle);
         reselectStream(m_currentAudioStream, MediaType::Audio);
         startCodecsLoading([=] {
-          mpv::qt::command(m_mpv, QStringList() << "hook-ack" << resumeId);
+          QLOG_INFO() << "resuming loading";
```
sixones: Nit, but maybe `MPV loading has been resumed`.

src/player/PlayerQuickItem.cpp
```diff
@@ -13,22 +15,16 @@
...
+#include "Globals.h"
 
 #if defined(Q_OS_WIN32)
 #include <windows.h>
 #include <dwmapi.h>
 #include <avrt.h>
 #endif
 
-#ifdef USE_X11EXTRAS
+#if defined(USE_X11EXTRAS)
```
sixones: Use `#ifdef`.

src/player/PlayerQuickItem.cpp
```diff
@@ -104,19 +91,38 @@ bool PlayerRenderer::init()
   DwmEnableMMCSS(TRUE);
 #endif
 
-  mpv_opengl_cb_set_update_callback(m_mpvGL, on_update, (void *)this);
+  static mpv_opengl_init_params opengl_init_params = {
```
sixones: `{` on next line.

src/player/PlayerQuickItem.cpp
```diff
@@ -104,19 +91,38 @@ bool PlayerRenderer::init()
...
+      get_proc_address, // .get_proc_address
+  };
+
+
+  static mpv_render_param params[] = {
```
sixones: `{` on next line.

src/player/PlayerQuickItem.cpp
```diff
@@ -104,19 +91,38 @@ bool PlayerRenderer::init()
...
+      {MPV_RENDER_PARAM_API_TYPE, (void *)MPV_RENDER_API_TYPE_OPENGL},
```
sixones: Space between `{` and `}`, same on the below lines.

src/player/PlayerQuickItem.cpp
```diff
@@ -158,9 +165,23 @@ void PlayerRenderer::render()
     }
   }
 
-  // The negative height signals to mpv that the video should be flipped
-  // (according to the flipped OpenGL coordinate system).
-  mpv_opengl_cb_draw(m_mpvGL, fbo, fboSize.width(), (flip ? -1 : 1) * fboSize.height());
+  mpv_opengl_fbo opengl_fbo = {
```
sixones: `{` on next line.

src/player/PlayerQuickItem.cpp
```diff
@@ -158,9 +165,23 @@ void PlayerRenderer::render()
...
+      fbo,              // .fbo
+      fboSize.width(),  // .w
+      fboSize.height(), // .h
+  };
+
+  mpv_render_param params[] = {
```
sixones: `{` on next line.

src/player/PlayerQuickItem.cpp
```diff
@@ -158,9 +165,23 @@ void PlayerRenderer::render()
...
+      {MPV_RENDER_PARAM_OPENGL_FBO, &(opengl_fbo)},
```
sixones: Spaces between `{` and `}`.

src/player/PlayerQuickItem.cpp
```diff
@@ -158,9 +165,23 @@ void PlayerRenderer::render()
...
+      {MPV_RENDER_PARAM_FLIP_Y, &flip},
+      {(mpv_render_param_type)0}
+  };
+
+  int ret = mpv_render_context_render(m_mpv_gl, params);
+  if (ret)
+  {
+      QLOG_ERROR() << "MPV Rendering reported an error " << ret;
```
sixones: Wrong indentation.

src/player/PlayerQuickItem.cpp
```diff
@@ -104,19 +91,38 @@ bool PlayerRenderer::init()
...
+  };
+
+
+  static mpv_render_param params[] = {
+      {MPV_RENDER_PARAM_API_TYPE, (void *)MPV_RENDER_API_TYPE_OPENGL},
+      {MPV_RENDER_PARAM_OPENGL_INIT_PARAMS, &opengl_init_params},
+#if defined(USE_X11EXTRAS)
+      {MPV_RENDER_PARAM_X11_DISPLAY, QX11Info::display()},
+#endif
+      {(mpv_render_param_type)0}
```
sixones: Use `nullptr` instead of `0`.

src/player/PlayerQuickItem.cpp
```diff
@@ -104,19 +91,38 @@ bool PlayerRenderer::init()
   DwmEnableMMCSS(TRUE);
 #endif
 
-  mpv_opengl_cb_set_update_callback(m_mpvGL, on_update, (void *)this);
+  static mpv_opengl_init_params opengl_init_params = {
+      get_proc_address, // .get_proc_address
```
sixones: Whats this comment for, seems like it's not needed?

sixones (approved):
Looks good, left some minor nitpicks and formatting changes. :shipit:

rcombs:
Why is the conanfile binary checked in?

src/player/PlayerQuickItem.cpp
rcombs: Indicates which member, since C++ doesn't have C-style designated initializers

src/player/PlayerComponent.cpp
rcombs: Is that the standard in PMP? We switched to `!` in PMS a while back.

src/player/PlayerQuickItem.cpp
rcombs: It's not a pointer?

rcombs: Though I would use `MPV_RENDER_PARAM_INVALID` here.

src/player/PlayerQuickItem.cpp
```diff
@@ -127,7 +133,8 @@ void PlayerRenderer::render()
 
   GLint fbo = 0;
   context->functions()->glGetIntegerv(GL_FRAMEBUFFER_BINDING, &fbo);
-  bool flip = true;
+  static bool flip = true;
```
rcombs: Why is this static now?

src/player/PlayerComponent.cpp
LongChair: No we use `!` everywhere else

src/player/PlayerQuickItem.cpp
LongChair: No it's not a pointer, just and end marker for the list.

src/player/PlayerQuickItem.cpp
LongChair: Yeah, the windows compiler won't allow to use C++11 initializers, so those are the struct field names.

src/player/PlayerQuickItem.cpp
LongChair: The parameters we pass in that struct are pointers, and the reason why mpv wants pointers is that mpv is making a copy of that struct internally.
So we want that variable to be persistent, otherwise will be lost on the copy and point to whatever is currently on the stack, flipping or not the image :)

LongChair: This is a recent change in mpv that might have been done after our update.
when i tried it with a more recent codebase, the screen would flip around like mad, which is why i made that change.

rcombs: Huh.
Could you move it to a class member instead, then?

LongChair:
Updated to PR according to comments

src/player/PlayerQuickItem.cpp
LongChair: I guess so if you feel that's important, we don't use that anywhere else though ...

rcombs: We also don't usually use mid-function statics.

LongChair:
@rcombs : Moved flip to class member and removed conanfile.pyo which should never have been there.

src/player/PlayerComponent.cpp
```diff
@@ -163,6 +163,11 @@ bool PlayerComponent::componentInitialize()
   mpv::qt::set_property(m_mpv, "hr-seek", "no");
   // Force vo_rpi to fullscreen.
   mpv::qt::set_property(m_mpv, "fullscreen", true);
+  // wait for cache to be filled before starting playback
+  mpv::qt::set_property(m_mpv, "cache-pause-wait", "yes");
```
rcombs: These two feel kinda off-topic; should they be in a separate commit?

LongChair: This is something that allows to wait until cache is filled.
As it was part of the new mpv features, i thought it would be ok for one line :)

src/player/PlayerComponent.cpp
sixones: The PMS spec still says `== false`, it does look like PMP is always using `!` so ignore me.

src/player/PlayerQuickItem.cpp
sixones: Isn't it obvious that `get_proc_address` == `.get_proc_address`, maybe I'm missing something?

LongChair: Well it turns out that our function has the same name as the struct property, but those could be different names. as we can't specify which member gets filled, and that it's only determined by the order or the params, so could as well be `foo`barlue going into `.bar` member.
We added that as well for all the other struct initializers.

Mitzsch01:
Anything new on this post? 
mpv seams to be pretty old as of today.

HazCod:
Peeking in, since the iOS/tvOS apps get mpv goodness.

mseeley:
@LongChair hasn't participating actively for a while and unfortunately this change got lost in the din.

I'll take this and push it over the finish line in January.

Mitzsch01:
Another question I have about this contribution is if it would be possible to change `vo=libmpv` to `vo=gpu`.
This would solve the vaapi hw decoding issue, as well as situations, were mpv.conf gets ignored. (especially options like `hwdec`, just my own experience with `vo=libmpv`) 
I tried setting it to vo=gpu myself, but it is not working correctly. One giant window outside of PMP gets rendered, without any option to interact with it (no GUI), but hwdec vaapi works.
I don´t know if this even possible to change the `vo` to `gpu` , but it would be nice.

mseeley:
HI @Mitzsch01. No, I'm not planning to make any changes aside from the minor code review feedback.

src/player/PlayerComponent.cpp
```diff
@@ -164,6 +163,11 @@ bool PlayerComponent::componentInitialize()
...
+  // define maximum cache time in seconds
+  mpv::qt::set_property(m_mpv, "cache-pause-secs", 20);
```
rcombs: This isn't a real property, afaik has never been. Not sure what was intended here; if it was `cache-secs`, that defaults to something much higher, which is generally desirable.

src/player/PlayerComponent.cpp
rcombs: `cache-pause-wait` is a number; this probably wants `cache-pause-initial`.

src/player/PlayerComponent.cpp
```diff
@@ -113,9 +113,8 @@ bool PlayerComponent::componentInitialize()
 
   // Do not let the decoder downmix (better customization for us).
   mpv::qt::set_property(m_mpv, "ad-lavc-downmix", false);
-
-  // Make it load the hwdec interop, so hwdec can be enabled at runtime.
-  mpv::qt::set_property(m_mpv, "hwdec-preload", "auto");
+ // Make it load the hwdec interop, so hwdec can be enabled at runtime.
```
rcombs: Misaligned.

rcombs (changes requested):

Mitzsch01:
> HI @Mitzsch01. No, I'm not planning to make any changes aside from the minor code review feedback.

Okay, that's fine. :)

HazCod:
> > HI @Mitzsch01. No, I'm not planning to make any changes aside from the minor code review feedback.
> 
> Okay, that's fine. :)

Open up a new issue @Mitzsch01

Mitzsch01:
@HazCod Not a bad idea, but I will wait until the main contribution gets merged. After that, I will take a look at it myself and will open an issue if needed.

src/player/PlayerComponent.cpp
mseeley: I think you're right @rcombs; I'll make that change.

src/player/PlayerComponent.cpp
mseeley: I think @LongChair may have intended to line up `cache-secs` and `cache-pause-wait` due to the recommendation of the doc above. I see we don't define `cache-secs` anywhere.

I'm planning to remove `cache-pause-secs` and see how our testers fair with remote playback.

src/player/PlayerQuickItem.cpp
mseeley: I see another usage of `{ MPV_RENDER_PARAM_INVALID }` in a similar params object. I'll mimic.

mseeley:
I've pushed testing builds of this PR to https://drive.google.com/drive/folders/1oQrTr7oPIDSwiObEld8b-Cn_-o-qJfj8?usp=sharing.

Mitzsch01:
I can even reproduce the issue with inverted/upside down playback on my 16:9 (3200x1800) Laptop. :/

Mitzsch01:
okay... :( 
Jep. the term "1.2" is very misleading.

After an additional reboot, the test build works fine. On my main rig (1920x1080), it also works fine, no upside down or inverted image. very strange
 

A small code addition. :) In the mpv logs (created by `mpv.conf` ` log-file=`)  one warning line gets printed.
`[  15.391][w][cplayer] Warning: property 'vo-drop-frame-count' was replaced with 'frame-drop-count' and might be removed in the future.`
This can easily be fixed, by changing the line 
 `info << "Dropped frames: " << MPV_PROPERTY("vo-drop-frame-count") << endl; `
to
` info << "Dropped frames: " << MPV_PROPERTY("frame-drop-count") << endl;`

mseeley:
@Mitzsch01 thank you for the feedback and fixup. I made the change in 3cbfec4 and have updated the builds on Drive; `2.24.0.936-3cbfec45` is the latest.

> After an additional reboot, the test build works fine. On my main rig (1920x1080), it also works fine, no upside down or inverted playback. very strange

This really feels like there's some incompatible state carried over from our production PMP builds and this test build. @dan-the-man-nz also mentioned the problem sorted itself out.

If you're feeling adventurous and reinstall the production app, play a video, replace the production app with this test build, then play the same video we'd know for sure. If you see the problem please hop into the TV layout and press `i`. You should see debug information like so:

screen shot 2018-12-21 at 11 03 46 am

We'd then know more about the graphics setup that leads to the flipping. The Desktop layout will support the same `i` behavior once I finish testing in #799.

We'll also be testing this ourselves to be sure but I bring it up in case you all have time and interest.

Thanks!

gbooker:
> The MPV changes in this branch did not cleanly build with our own LibreELEC 8.2.2 dependency. We had to bump to LibreELEC 8.2.5. Upgrading is clearly a favorable step but we need to first discuss.

For the non-plex contributors, we later discovered that this is not to do with LE 8.2.2 vs 8.2.5 but rather a commit present on our 8.2.5 branch that bumps the MPV version in LE.  Our 8.2.2 branch uses MPV from the branch `old-master` in our MPV repo which is the same MPV we were building PMP against (until this PR that is).  Our 8.2.5 branch uses MPV from the branch `v1.2` which is what the PR changes PMP to use.

So, with this in mind, there are four combinations:

MPV (matching LE and PMP) | `old-master` | `v1.2`
------ | --- | ------
LE 8.2.2 | 1 | 2
LE 8.2.5 | 3 | 4

1. Our previous builds of LE internal and those given to the public
1. Untested (AFAIK)
1. I personally have tested this several months ago when I was playing with internal PMP versions.  This did work (and AFAIK still does).
1. I'm currently running this on my NUC using the PMP branch in this PR.

(2) may require some adjustments of other dependencies in that LE branch, perhaps the FFmpeg version, or it may just work outright.  I've not spent the ~6 hours it takes to build it to find out.

dan-the-man-nz:
> @Mitzsch01 thank you for the feedback and fixup. I made the change in [3cbfec4](https://github.com/plexinc/plex-media-player/commit/3cbfec45da806b896a4ae998d2f188d33316dd2a) and have updated the builds on Drive; `2.24.0.936-3cbfec45` is the latest.
> 
> > After an additional reboot, the test build works fine. On my main rig (1920x1080), it also works fine, no upside down or inverted playback. very strange
> 
> This really feels like there's some incompatible state carried over from our production PMP builds and this test build. @dan-the-man-nz also mentioned the problem sorted itself out.
> 
> If you're feeling adventurous and reinstall the production app, play a video, replace the production app with this test build, then play the same video we'd know for sure. If you see the problem please hop into the TV layout and press `i`. You should see debug information like so:
> 
> screen shot 2018-12-21 at 11 03 46 am
> 
> We'd then know more about the graphics setup that leads to the flipping. The Desktop layout will support the same `i` behavior once I finish testing in #799.
> 
> We'll also be testing this ourselves to be sure but I bring it up in case you all have time and interest.
> 
> Thanks!

Hi
on another HTPC, single tv attached, i installed PlexMediaPlayer-2.24.0.936-3cbfec45-windows-x64.exe test release.

first time playing a video, it was upside down
[PlexMediaPlayer.log](https://github.com/plexinc/plex-media-player/files/2704330/PlexMediaPlayer.log)

image
image

then i closed the PMP app, reopened, and it was back to normal - playback was no longer upside down.
(same video)
[PlexMediaPlayer.log](https://github.com/plexinc/plex-media-player/files/2704341/PlexMediaPlayer.log)

Mitzsch01:
I'm not able to reproduce the upside down playback on my main rig nor on my laptop. (and no dropped frames)
Laptop: (which has shown this strange behavior)


And my main rig (AMD graphics)

Mitzsch01:
Another thing I have encountered while testing. HDR seems muted, no rich color. Since I have used a self-compiled Version of PMP on Linux (including this PR here), Qt 5.11.2 and also with the latest FFmpeg/ mpv. On Linux the same HDR file looks gorgeous. I´ve made a comparison picture. (left - `2.24.0.936-3cbfec45` , right - self compiled)
mpv_comparison_2_1
 
Info overlay of the Linux machine. (I have modified the overlay a bit, so ffmpeg and mpv version is exposed right away) 
bildschirmfoto vom 2018-12-22 15-02-48_blur

gbooker:
@Mitzsch01
- Are you advocating skipping the `v1.2` branch and going straight to the current 0.29.1 version?
- Is the QT version bump necessary? (trying to localize changes to smaller incremental steps)
- Is the FFmpeg bump necessary? (this requires some additional fun on our side)

Mitzsch01:
> Are you advocating skipping the v1.2 branch and going straight to the current 0.29.1 version?

I think it´s a small miss understanding on my side. Does `v1.2` points to a specific mpv (28?) version? 
Or does it means a refresh of existing pmp code?
My question would be why upgrading to an already old version? 
Why not bump to the latest version? (29.1, as it brings many new features or bug fixes) 
It´s working, especially with this PR here, (like the move from opengl-cb to libmpv)

> Is the QT version bump necessary? (trying to localize changes to smaller incremental steps)

I thought it´s already planned? Would also improve performance. (on my Linux machine Qt 5.11.2 seems faster and more reliable than Qt 5.9.5, no weirdness with dropped frames)
(https://github.com/plexinc/plex-media-player/issues/802)

> Is the FFmpeg bump necessary? (this requires some additional fun on our side)

I thought it´s already planned?  It´s working already. ;) (and needed if bumping to a newer mpv version)
(https://github.com/plexinc/plex-media-player/issues/820)

lqvnguyen:
Another issue with bumping mpv past the current 0.26, is that deinterlacing has changed dramatically.
The commit that changed it:
https://github.com/mpv-player/mpv/commit/830f0aed97da7257e1cbe30a367144bc51ec52d2
Detailed Notes re: all the changes since 0.26
https://github.com/mpv-player/mpv/blob/master/DOCS/interface-changes.rst

```
--- mpv 0.27.0 ---
   - remove that hardware deinterlace filters (vavpp, d3d11vpp, vdpaupp)
     changed their deinterlacing-enabled setting depending on what the
     --deinterlace option or property was set to. Now, a filter always does
     what its filter options and defaults imply. The --deinterlace option and
     property strictly add/remove its own filters. For example, if you run
     "mpv --vf=vavpp --deinterlace=yes", this will insert another, redundant
     filter, which is probably not what you want. For toggling a deinterlace
     filter manually, use the "vf toggle" command, and do not set the
     deinterlace option/property. To customize the filter that will be
     inserted automatically, use --vf-defaults. Details how this works will
     probably change in the future.
   - remove deinterlace=auto (this was not deprecated, but had only a very
     obscure use that stopped working with the change above. It was also
     prone to be confused with a feature not implemented by it: auto did _not_
     mean that deinterlacing was enabled on demand.)
```

gbooker:
> I think it´s a small miss understanding on my side. Does v1.2 points to a specific mpv (28?) version?

I'll point to [an earlier comment](https://github.com/plexinc/plex-media-player/pull/727#issuecomment-449219073):
> "v1.2" is a branch name from our MPV fork. It's entirely misleading. In reality this is a version revision of MPV 0.28 (mpv-x86_64-20180421-git-65f0825).

> My question would be why upgrading to an already old version?

This is why I asked about whether the FFmpeg bump is necessary.

> I thought [FFmpeg/QT bumps are] already planned?

The issue is whether a bump in MPV is dependent on these.  If it is, then it delays it until the work on those dependencies is done.  If a bump is not dependent, then incremental progress can be made here before/independent of the dependencies work.  Additionally we must ask how much work should be done in dependencies in this project prior to upgrading the dependency build system to what we are using internally on other projects?  There are potentially a lot of moving pieces and if the changes are localized to a smaller set, they are easier to make.

Honestly, I think the biggest issues are the playback regressions.  These include the video flipping and the implications on deinterlace, where the latter doesn't go away with a newer MPV.

Mitzsch01:
> "v1.2" is a branch name from our MPV fork. It's entirely misleading. In reality this is a version revision of MPV 0.28 (mpv-x86_64-20180421-git-65f0825).

Oh, lol I missed that. Allthough i´ve read this, xD I´m sorry. ^^

-----
Newer mpv versions need FFmpeg 4+
-> https://github.com/mpv-player/mpv/releases/tag/v0.29.0 

> This release requires FFmpeg 4.0 or newer.


Qt (5.11.2) does not require a newer mpv/FFmpeg version. 

> Honestly, I think the biggest issues are the playback regressions. These include the video flipping and the implications on deinterlace, where the latter doesn't go away with a newer MPV.

Well, that's true, video flipping is a really bad thing. 
(but is it entirely related to this new code? (because on Linux the PR is working great) or is it due to an old opengl/angle version?) 
In my opinion its worth the hassle, because future upgrade/ new features can be implemented easily and the playback seems more robust. (except flipping ^^) 
In my opinion its totally worth the hassle, because future upgrade/ new features can be implemented easily and the playback seems more robust. (except flipping ^^) But I have no idea what has to be changed on your side and how much work is needed (eg dependencies as you said) to accomplish the goal of this PR.

lqvnguyen:
> In my opinion its totally worth the hassle, because future upgrades/ new features can be implemented easily and the playback seems more robust. (except flipping ^^) But I have no idea what has to be changed on your side and how much work is needed (eg dependencies, as you said) to accomplish the goal of this PR.
Have you been able to overcome HW decoding limited to vaapi-copy?

Mitzsch01:
> Have you been able to overcome HW decoding limited to vaapi-copy?

Thanks for asking. :) Unfortunately no, but I´ve not looked into details yet. 
But as far as I can tell only `vo=gpu` will work. `vo=libmpv` has some strange limitations, I have absolutely no idea. Even in other applications like GNOME MPV and `vo `set to `libmpv`, vaapi (non-copy) is not working.
`vo=gpu` seems more convenient. That's why I´ve asked to change the code. https://github.com/plexinc/plex-media-player/pull/727#issuecomment-447963734

mseeley:
I haven't had a good handling on `vo=` values. I found this handy doc: https://github.com/mpv-player/mpv/wiki/Which-VO-should-I-use%3F.

Changing to `vo=gpu` as the default value would be interesting. I'm unsure what would break and how since it seems to be scarily dependent on the GPU driver implementations.

derzwen:
Hi, in the past 2 weeks, i've played around with this PR because I had some serious problems with the latest plex player >2.20 running on my Intel Skylake hardware with Ubuntu 18.10.

With >2.20 using Knapsu's images, I had the following problems on Ubuntu 18.10
- HW decoding was not working
- Audio passthrough was not working (not a problem this PR solves)

So I basically looked at Knapsu's work, understood most part of it and built my own AppImage using a Docker container and his scripts (both with slight modifications). As of right now, I've a working AppImage for the latest PlexPlayer master having full HW decoding (not only copy) and audio passthrough working on my Intel maschine. This AppImage uses the latest Plex Player master and  latest MPV/FFMPEG master during AppImage build.

I did lot's of try and error and I want to share some observations. I got full HW decoding working in ubuntu 18.10 at least on intel via vaapi-egl with the following mpv settings combination:

**plexmediaplayer.conf**
```
debug.force_vo="gpu"         (so use vo=gpu instead of vo=libmpv)
```

**mpv.conf**  (all are optional, they will be correctly auto probed by "gpu" at least on my intel maschine)
```
gpu-context=x11egl
gpu-hwdec-interop=vaapi-egl
hwdec=vaapi
```

If I don't force the vo to "gpu" (and use "libmpv" instead), the other 3 properties in mpv.conf get ignored (and auto probing does not happen) and HW decoding does not work (at least as of right now, latest mpv master) That's the same problem i think @Mitzsch01 mentioned 7 days ago. (issue [mpv-player/mpv#6353](https://github.com/mpv-player/mpv/issues/6353)) If this is really a bug on the MPV side, "libmpv" as vo might be sufficient in the future. As of right now, full hw decoding for me works only if I use "gpu" so that the other settings don't get ignored.

Can I help any further? Should I post more details on how I build the AppImage?

I've attached to logs from PlayerComponent playing a standard H264 movie doing probing with "gpu" and using "libmpv" and those 3 settings mentioned earlier. "gpu" uses full hw decoding, "libmpv" falls back to copy acceleration because the mpv.conf settings are ignored.

[gpu.txt](https://github.com/plexinc/plex-media-player/files/2735558/gpu.txt)
[libmpv.txt](https://github.com/plexinc/plex-media-player/files/2735559/libmpv.txt)

mseeley:
Hi @derzwen 👋  thanks for the details!

We got a little scared off `vo` of `gpu` based on one of the posts above: https://github.com/plexinc/plex-media-player/pull/727#issuecomment-447963734. I wonder what's different. Regardless we've got some interesting stuff to sort through. :)

Mitzsch01:
@derzwen 

> That's the same problem i think @Mitzsch01 mentioned 7 days ago.

yes, exactly the same! 

> "libmpv" as vo might be sufficient in the future.

or obsolete, but it´s the same as I think. 



` "debug.force_vo": "",` is quite handy to have

with  `"debug.force_vo": "gpu",` I got this 


As I said earlier "no way to interact with it", well that`s not really the case. The controls are not visible because of the windows, but you can resize it and see the PMP window, with playback controls. They work.

as well as vaapi hw decoding

rcombs:
We won't be switching the vo used. The `gpu` vo is not appropriate for the application embedding that PMP uses; that's what the `libmpv` vo (previously known as `opengl-cb`) is for. They share the vast majority of their code; if there's a bug resulting in problems with hardware acceleration under the libmpv vo, we can gather information on what's causing it and figure out a solution. Is that problem a regression in this PR for you? If not, file a separate GitHub issue and move this discussion there.

Mitzsch01:
You are right but there are applications out there that seem to embed mpv with vo=gpu. So it should be possible even when it´s not designed to be used like that. e.g. Emby Theater is such a candidate. There are some pictures of the info overlay of emby theater online, that indicate vo=gpu. 
Solving this libmpv hw decoding bug would be nice but is it worth the hassle or should we investigate in changing the code so vo=gpu is possible? :) 

> Is that problem a regression in this PR for you? 

nope, even with opengl-cb I had issues on Linux. (AMD GPU system) :/

derzwen:
At least there's is the possibility so use a different vo than libmpv for the more advantageous guys out there using debug.force_vo. So that's good! And maybe also sufficient right now until this get's fixed by the MPV guys.

It would be really nice to get this PR through to have PMP use a recent MPV version even with vo=libmpv and not an old 0.28.xxx git custom forked one.

Concerning the "off" MVP window: I've never had those problems on my machine. I have also never experienced this flipped content problem mentioned earlier. But maybe because I'm running all of this on Ubuntu 18.10 with QT 5.11.2 + Ubuntu 18.10 also used for building and bundling the AppImage.

Can we help any further in terms of testing to get this PR ready?

rcombs:
Please move discussion that isn't about regressions or testing of this PR in particular to separate GHIs. Once everything's moved out, I'll likely be deleting old irrelevant comments (including this one) to make this thread readable so we can actually manage this change. I'll remind here again in a week, and then delete regardless.

Mitzsch01:
Ok then I will open a new issue regarding changing the `vo` or better to get `vaapi` working.

> flipped content problem

this was only present on windows, Linux was not affected, luckily.

HazCod:
Peeping in again

mseeley:
I'm closing this for now. We'll reevaluate in the future.

Mitzsch01:
Any chance that this get´s revisited? (Now that PMP isn´t EOL anymore)

gbooker:
Internally we are experimenting with making a build of PMP using MPV 0.28.2 and FFmpeg 1.6 which supplants the work in this PR.

LongChair:
This PR is most likely not much bleeding anymore :)

gbooker:
https://github.com/plexinc/plex-media-server-ffmpeg-gpl/tree/v1.6

rcombs:
Sorry about the confusion; for clarification, this refers to our internal versioning, which is not related to upstream ffmpeg versions (which we don't track).

Mitzsch01:
Ah!, okay, that makes sense. ^^
But is there an equivalent FFmpeg upstream version or how recent is "FFmpeg 1.6"?

PR #757, opened 2018-06-23
Display Rate Fixes

gbooker:
Rationale:
- Have a TV that supports both 59.94Hz and 60Hz display rates
- Have display currently on 60Hz
- Play content that is 29.97Hz
- Prior to this change, the 60Hz rate was within 0.1Hz of 2*29.97Hz so it would be deemed an exact multiple match and content would play at 60Hz
- After this change, 59.94Hz would be deemed an exact multiple match (properly) and 60Hz would be deemed a close multiple match (properly).
- This change also mirrors the exact match logic which uses a tolerance of 0.01 where as the multiple match was previously using a tolerance of 0.1.  Since there is a multiplicative factor involved in this function, I multiplied 0.01 by that factor to achieve the final tolerance.

### DIFF
--- a/src/display/DisplayManager.cpp
+++ b/src/display/DisplayManager.cpp
@@ -85,11 +85,12 @@ bool DisplayManager::isRateMultipleOf(float refresh, float multiple, bool exact)
   if (roundedRefresh == 0)
       return false;
 
-  float newRate = roundedMultiple / roundedRefresh * refresh;
+  long factor = roundedMultiple / roundedRefresh;
+  float newRate = factor * refresh;
   if (newRate < 1)
     return false;
 
-  float tolerance = exact ? 0.1 : 1;
+  float tolerance = exact ? 0.01 * factor : 1;
 
   return fabs(newRate - multiple) < tolerance;
 }


LongChair:
The change is small.
Did you check if it would pick the right ones on the close refresh rates, ie a scenario where you have 23.975 iirc vs 24 and also 29.9x vs 30 hz ?

otherwise LGTM

Moussa:
@mseeley Thanks for the poke. This should definitely be ported over to TV's `getBestCandidateDisplayMode` 👍

gbooker:
I can personally confirm the following with this diff:

content refresh rate | selected refresh rate | notes
--- | -- | --
23.976 | 23.976 | None
29.97 | 29.97 | `refreshrate.avoid_25hz_30hz` is `false`
29.97 | 59.94 | `refreshrate.avoid_25hz_30hz` is `true`
25 | 50 | None

TV supports: 23.976, 24, 29.97, 30, 50, 59.94, and 60 (all progressive).  I believe that covers all the test cases you wanted.

Prior to this diff, the 29.97 content was selecting 60 instead of 59.94 when `refreshrate.avoid_25hz_30hz` was set (3rd line in table).

LongChair:
LGTM

lqvnguyen:
I agree with the commits

HazCod:
Thank you for this PR! I had the exact same issue.

PR #778, opened 2018-09-14
Background power consumption optimizations

rcombs:
I've got one other tweak that improves this further, but needs some more work before merge. These should be fairly safe.


HazCod:
@rcombs any plans to merge this?

mseeley:
Hi @HazCod thanks for the ping. We'll stoke this one again.

Issue #743, opened 2018-05-26
player: PMP won't start, generating large .xsessions-errors file


wamari:
### Test environment
PMS Version: N/A
PMP Version: 2.10.0.849-e02dbeca
Platform : Antergos Linux

### Steps to reproduce
1. Installed latest version of PMP, it may also be an issue with qt5.11

### Current behavior
1. PMP stuck on splash screen
2. Genrating large .xsession-errors file, in my case upto 90GB spewing the following lines
```
[17938:18165:0526/232935.585345:ERROR:gl_surface_qt.cpp(305)] eglCreatePbufferSurface failed and surfaceless context not available
[17938:18165:0526/232935.585356:ERROR:gl_surface_qt.cpp(312)] Requested OpenGL implementation is not supported. Implementation: 0
```

### Expected behavior
1. PMP should open normally

erenfro:
I am having this issue as well. Soon as you click on a user in a multi-user setup in Plex, those errors start spewing out. This only started happening after updates were installed, of which qt 5.11 was included in that from 5.10.

wamari:
Have you tried downgrading Qt 5.11? I tried but still not working...

erenfro:
I managed to get this working again by downgrading back to qt-5.10 from my package caches, and re-compiling against 5.10 again, so yes, it's some new problem spawned from 5.11 specifically.

wamari:
i also downgraded my Qt 5.11 packages to Qt 5.10 and recompiled, working great now...

ludenticus:
Why was closed this issue?

mannp:
@LongChair shouldn't this be opened until plex catches up with the latest qt builds?

erenfro:
I kinda have to agree in the sense of the actual problem, why it was reported, and the fact  that it does effect PMP now, and in the future.

erenfro:
Also, for reference. Solus issues regarding this as well:
https://dev.solus-project.com/T6542

AndreasBackx:
@LongChair any update or ETA on this? Downgrading can be a pita on Arch.

ATRIKO:
Completely unusable on my Arch Linux HTPC without qt 5.11 support.

I've canceled my Plex Pass until you have qt 5.11 support.

mseeley:
Longchair is out for a bit, bumping so we see this during triage.

wamari:
Completely unusable for now with qt 5.11 resorted to using kodi with plex add on until this is resolved.

[...]

PR #785, opened 2018-10-12
Update README instructions regarding Qt versions

jleroy:
Since version 5.11.2, Qt 5.11 is working well with PMP. I've updated the
README to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Or if you prefer working in Xcode, run CMake for the xcode build):
 
 ## Building on Linux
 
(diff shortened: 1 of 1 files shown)


mseeley:
Hi @jleroy thank you for the doc update. Before landing we need to figure a couple things

- What the Qt fix was between 5.11.1 and 5.11.2? I think it'll help build our confidence around the documentation update if we can correlate build support to an intentional change in Qt.
- I've asked any interested Linux user in #743 for another 👍 that 5.11.2 builds correctly. Signs definitely point to "yes" but Qt compat has been a particularly troublesome issue.

Again, thanks!

rphillips:
Upstream QT Bug: https://bugreports.qt.io/browse/QTBUG-68479
Upstream QT Patch: https://codereview.qt-project.org/#/c/232017/

jleroy:
@mseeley: @rphillips answer was the good one. :)

mseeley:
Excellent, thank you both! Let's :shipit:

PR #873, opened 2019-07-09
GTK dark titlebar

mxre:
Set X11 Window Property for GTK darker titlebar. This pure cosmetic
patch makes the titlebar less intrusive when using the player in
windowed mode.

Plex-CLA-1.0-signed-off-by: Max Resch <resch.max@gmail.com>

### DIFF
--- a/src/ui/KonvergoWindow.cpp
+++ b/src/ui/KonvergoWindow.cpp
@@ -21,6 +21,11 @@
 #include "Globals.h"
 #include "EventFilter.h"
 
+#ifdef Q_OS_UNIX
+#include <QX11Info>
+#include <X11/Xlib.h> 
+#endif
+
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 class ScopedDecrementer
 {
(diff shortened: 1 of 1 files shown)


mseeley:
Thanks @mxre. 😄 

Does this cause the title bar to always be dark on supporting systems? If so I have some reservation as we'd be changing behavior for folks that enjoy the lighter titlebar.

@rcombs will you please review the implementation?

mxre:
Yes, this would cause the titlebar to be always dark (at least on Gnome3 and maybe other supported systems). This would match the default Plex theme which is also in a dark tone.

For reference I enclose the following images, comparing light and dark titlebars in the default theme Adwaita.




It is also true that not all gtk themes have an alternate mode. So in these cases nothing would change. I argue, that using the darker theme alternative for an application that already uses a dark theme, would not only match plex' overall look but would also be consistent with other applications like eog or gnome-mpv.

Maybe it is a my personal taste, I just wanted to start a conversation on this.

I could amend my commit to include a configuration option in the "main " section like "preferGtkDark": true

Issue #952, opened 2020-03-31
No audio at all on arm64 and amd64 current 2.54-dev


mattgorski:
**PMP Version: 2.54-dev
**Platform Linux/Ubuntu 18.04 amd64 Laptop and NVIDIA Jetson TX1 arm64

### Steps to reproduce
1.  Followed build instructions with system QT v5.9.5 ubuntu packages

### Current behavior
1. Zero sound (AC3) even after testing all device outputs in Audio Settings.  When pressing "i" info on keyboard nothing shows for audio output/codec/driver etc.

Built MPV on both systems and audio works fine.  Tried custom mpv.conf with same results.

### Expected behavior
1. Sound

mseeley:
Logs?

mattgorski:
@mseeley Grabbing both logs in a min.  Thanks for the reply :)

mattgorski:
Here is the log from my NVIDIA Jetson TX1 (which I am mainly focused on for a pmp system)

https://pastebin.com/uChKmUj5

mseeley:
@gbooker, @rcombs do you see anything out of the ordinary with the audio device logs above?

chainikdn:
Here's the mpv change removed legacy hooks API: https://github.com/mpv-player/mpv/commit/2337fa4e0213993398d36cb3222633766d677dfd

mattgorski:
Thanks a million guys I'll try to build mpv with the provided changes.

kennysgithub:
FWIW this also affects x86_64 builds (Kubuntu 19.10). Winding back my mpv to commit `9abb710a` didn't fix the issue for me, however. I may try winding mpv back to `2337fa4^` if I get time later today.

mattgorski:
Still no audio for me checking out mpv at 2337fa4.  I can hear click navigation sounds just no audio driver/codec etc even after selecting alsa/default hdmi 0.3.

kennysgithub:
Go up one more, to ` 2337fa4^` (the up-arrow means "the commit before that one").

mattgorski:
Ahhhh ok one up thanks @kennysgithub

kennysgithub:
BTW, should we be building PMP against the original MPV repo, or Plexinc's fork, or does it matter?

mattgorski:
[...]

PR #953, opened 2020-04-15
player: Use new hooks API for libmpv > 107

kennysgithub:
This should fix the problem listed in issue #952 . Seems to work OK (and has backwards API compatibility, but I haven't tested that yet).

### DIFF
--- a/src/player/PlayerComponent.cpp
+++ b/src/player/PlayerComponent.cpp
@@ -188,11 +188,15 @@ bool PlayerComponent::componentInitialize()
   // Setup a hook with the ID 1, which is run during the file is loaded.
   // Used to delay playback start for display framerate switching.
   // (See handler in handleMpvEvent() for details.)
-  mpv::qt::command(m_mpv, QStringList() << "hook-add" << "on_load" << "1" << "0");
-
   // Setup a hook with the ID 2, which is run at a certain stage during loading.
   // We use it to initialize stream selections and to probe the codecs.
+#if MPV_CLIENT_API_VERSION < MPV_MAKE_VERSION(1, 100)
+  mpv::qt::command(m_mpv, QStringList() << "hook-add" << "on_load" << "1" << "0");
   mpv::qt::command(m_mpv, QStringList() << "hook-add" << "on_preloaded" << "2" << "0");
+#else
+  mpv_hook_add(m_mpv, 1, "on_load", 0);
+  mpv_hook_add(m_mpv, 2, "on_preloaded", 0);
+#endif
 
   updateAudioDeviceList();
   setAudioConfiguration();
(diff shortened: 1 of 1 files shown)


src/player/PlayerComponent.cpp
```diff
@@ -648,6 +655,49 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
       }
       break;
     }
+    case MPV_EVENT_HOOK:
+      // FIXME: do these actually need to be filtered on old libmpv versions?
```
kennysgithub, line 659: Do these actually need to be filtered out on older libmpv versions? (IOW, will the old API generate both "message" events and "EVENT_HOOK" events?)

mseeley:
Hi @kennysgithub - thanks!

Please update your commit message according to this doc: https://github.com/plexinc/plex-media-player/blob/master/CONTRIBUTING.md

src/player/PlayerComponent.cpp
rcombs: The old message events are only generated for legacy hooks.

This block should be preprocessor-`#if`'d as well.

src/player/PlayerComponent.cpp
```diff
@@ -648,6 +655,49 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
...
+      if (mpv_client_api_version() >= MPV_MAKE_VERSION(1, 108))
+      {
+	mpv_event_hook *hook = (mpv_event_hook *)event->data;
```
rcombs, line 662: Since you're declaring a local here, you'll still need the block (just with no need for the runtime `if`); it should be indented at the same level as the `case`, as with the other cases.

kennysgithub:
I think I fubared() the set of changes (I'm used to the method used by Gerrit) so there may be a new commit, but I've made the changes requested.

src/player/PlayerComponent.cpp
```diff
@@ -188,11 +188,15 @@ bool PlayerComponent::componentInitialize()
   // Setup a hook with the ID 1, which is run during the file is loaded.
   // Used to delay playback start for display framerate switching.
   // (See handler in handleMpvEvent() for details.)
-  mpv::qt::command(m_mpv, QStringList() << "hook-add" << "on_load" << "1" << "0");
-
   // Setup a hook with the ID 2, which is run at a certain stage during loading.
   // We use it to initialize stream selections and to probe the codecs.
+#if MPV_CLIENT_API_VERSION < MPV_MAKE_VERSION(1, 108)
```
rcombs, line 193: Why 1.108? The legacy hook API was removed in 1.107, and the replacement was added in 1.100.

src/player/PlayerComponent.cpp
```diff
@@ -648,6 +652,50 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
       }
       break;
     }
+#if MPV_CLIENT_API_VERSION >= MPV_MAKE_VERSION(1, 108)
```
rcombs, line 655: Need to adjust this one as well.

src/player/PlayerComponent.cpp
```diff
@@ -648,6 +652,50 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
...
+	}
+	else
+	{
+	  resume();
+	}
+	break;
+      }
+      if (!strcmp(hook->name, "on_preloaded"))
+      {
+	reselectStream(m_currentSubtitleStream, MediaType::Subtitle);
```
rcombs, line 688: Last thing: please detab (looks like your editor did GNU-style mixed 8-wide tabs here).

src/player/PlayerComponent.cpp
```diff
@@ -648,6 +652,50 @@ void PlayerComponent::handleMpvEvent(mpv_event *event)
       }
       break;
     }
+#if MPV_CLIENT_API_VERSION >= MPV_MAKE_VERSION(1, 100)
```
rcombs, line 655: Oh, and you'll want your commit message to have a top line like "player: Use new hooks API for libmpv >= 1.100; fixes #952"

kennysgithub:
OMG, just take the patch; y'all can have it. I'm done

PR #957, opened 2020-04-23
Kill the helper on shutdown/reboot.

gbooker:
On linux systems, when the shutdown or reboot is triggered from PMP, the helper can suspend the shutdown/reboot process for 90s (default) while the OS is waiting for it to quit and not doing so.  This kills the helper before we get there and thus enabling a much faster shutdown/reboot in those circumstances.

### DIFF
--- a/src/power/PowerComponentDBus.cpp
+++ b/src/power/PowerComponentDBus.cpp
@@ -3,6 +3,7 @@
 #include <QtDBus/QDBusReply>
 
 #include "PowerComponentDBus.h"
+#include "utils/HelperLauncher.h"
 
 #define DBUS_SERVICE_NAME "org.freedesktop.login1"
 #define DBUS_SERVICE_PATH "/org/freedesktop/login1"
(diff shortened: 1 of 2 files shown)


gbooker:
I'm not sure of the exact circumstances leading to the helper not being quit during a reboot/shutdown but I can tell you it's not consistent.

In my setup I have Ubuntu 18.04 set to auto-login and upon login it'll launch PMP.  If I tell it to shutdown or reboot from PMP, sometimes it'll hang during the shutdown process waiting for the pmp helper to exit and if it encounters this, it always times out and kills it outright.
This occurs on two separate installs so it's not unique to one.  In both cases, the above code change means the shutdown is not impeded.
It's a very dramatic difference in that the shutdown speed goes from 93 seconds -> 3 seconds.

Mitzsch01:
Does the pmphelper even has a function under Linux? I mean you can compile pmp without pmphelper by adding `-DENABLE_HELPER=off` to the CMake line and you are good to go. (this is how I do it)

PR #962, opened 2020-05-01
Fix method invocation ambiguity

dschmidt:
This fixes 

```
/usr/include/qt5/QtCore/qlist.h:210:10: note: candidate function
    void append(const T &t);
         ^
/usr/include/qt5/QtCore/qlist.h:211:10: note: candidate function
    void append(const QList<T> &t);
```

with my Qt and compiler version.


mseeley:
Sorry for the incredibly long delay landing this. Thanks a lot for addressing!

